Issue #94, opened 2025-07-16
Refactor plugin architecture


adamlacoste:
Presently Specify's plugin architecture assumes that all plugins are merely bundles of Cucumber support code, and that including their (complexly resolved) source paths in Cucumber's `import` array is the appropriate way to handle them.  Not only is this approach somewhat complicated and brittle, but it drastically limits what Specify plugins can be used for in the future.  It would be better if we refactored the plugin architecture to designate Cucumber support code plugins as such, leaving room for other kinds of plugin behavior later.  As a bonus, doing this will allow each Cucumber plugin to provide its own Cucumber support code paths as an export, meaning that no complicated resolution logic will be needed any longer.

PR #106, opened 2025-07-16
Plugins refactored

adamlacoste:
[#94] Refactored how plugins are processed, shifting more responsibility to the plugin module to structure its default export and thereby removing a bunch of brittle module resolution logic in core

### DIFF
--- a/modules/@specify/core/src/lib/TestCommand.test.ts
+++ b/modules/@specify/core/src/lib/TestCommand.test.ts
@@ -1,7 +1,6 @@
-import merge             from "deepmerge";
-import { createRequire } from "module"; 
-import fs                from "node:fs";
-import path              from "node:path";
+import merge from "deepmerge";
+import fs    from "node:fs";
+import path  from "node:path";
 
 import {
     TestCommand,
@@ -10,7 +9,6 @@ import {
 } from "./TestCommand";
 
 describe("TestCommand", () => {
-    const require   = createRequire(import.meta.url);
     const emptyArgs = { "_": [] };
     const emptyOpts = {};
 
@@ -25,7 +23,6 @@ describe("TestCommand", () => {
 
                 expect(cmd.cucumber).toEqual(expOpts.cucumber);
                 expect(cmd.gherkinPaths).toEqual(expOpts.gherkinPaths);
-                expect(cmd.plugins).toEqual(expOpts.plugins);
             });
 
             it("...with merged values when user options are provided", () => {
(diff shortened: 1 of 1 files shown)


modules/@specify/core/src/lib/TestCommand.ts
```diff
@@ -6,7 +6,6 @@
  */
 
 import merge              from "deepmerge";
-import { createRequire  } from "module";
```
adamlacoste, line 9: So I guess file-level comments are gone now?  Alright, I'll say what I would have said there, here.

My refactor was able to remove _all_ plugin-related logic from `TestCommand`.  All that stuff is fully baked into the Cucumber config that gets passed in with the constructor's `opts` param at instantiation.  It's nice when things can be simpler.

modules/@specify/core/src/main.ts
```diff
@@ -22,19 +23,35 @@ import {
...
+            case "cucumber":
+                cucumberCfg = merge(cucumberCfg, mod.default.cucumber);
+                break;
+        }
+    }
+}
+
+// add core cucumber support code path
+cucumberCfg.import.push(path.resolve(import.meta.dirname, "cucumber"));
+
```
adamlacoste, line 44: Here's the new logic that processes plugin metadata and updates the Cucumber config accordingly.

For each plugin listed in our config -- which should work equally well for package names and paths -- we import the module and then loop through the properties of its **default** export.  (Named exports can therefore be used for anything else that a plugin author might want them to do, e.g. providing handy utilities for use in other places.  Presently, the only property we check for is "cucumber", and if found, we treat it as an extension of for the Cucumber config.

This has several benefits, the most notable being that we don't need to do a lot of processing here or in `TestCommand` to convert the plugin's data into Cucumber's preferred format.  But it also means that plugins can modify Cucumber's behavior in a variety of ways, putting more power in the hands of the plugin creator.

The last thing we do is ensure that the local "./cucumber" path is added to the Cucumber config's import list, because we need our core module's support code to be included at all times.  We could put this in the Cucumber config, but then there is a chance that users would override it and all kinds of things (like quick ref lookups) would break.

modules/@specify/plugin-cli/src/main.ts
```diff
@@ -5,3 +5,13 @@
  * support code facilitating tests which need to interact with a command-line
  * interface.
  */
+
+import path from "node:path";
+
+export default {
```
adamlacoste, line 11: Worth noting and discussing here is that the meta data this plugin is exporting is very minimal.  Originally I had imagined providing a lot more details (name, author, etc.) that would generally be redundant with the module's package.json, and I realized that our core logic has no use for any of that noise.  If anyone can think of more meta data we should be exporting, I'm open to suggestions, but my inclination for now is to keep it lean and augment it over time as more needs arise.

MentalFusion (approved):

wild-love (approved):

PR #109, opened 2025-07-18
Line width (linter fix)

adamlacoste:
[#108] Added 120 character line width setting to Prettier


MentalFusion (approved):

wild-love (approved):

.prettierrc.json
```diff
@@ -1,5 +1,6 @@
 {
     "endOfLine": "auto",
+    "printWidth": 120,
```
adamlacoste, line 3: Point of discussion: Prettier strongly recommends leaving this at 80 because it's not an enforced limit, it's a guideline that Prettier will attempt to target but has latitude to interpret.  I think 80 is too short and leads to too much unnecessary line-breaking, so I'm happy to go higher.  But is 120 the target number we should be shooting for, or should it be somewhere in between?

- [ ] Discussed `printWidth`

PR #111, opened 2025-07-22
Multi-shell init

adamlacoste:
`SessionManager` (formerly `Commander`) now juggles multiple shells in parallel.  Pre-existing methods all assume they should use the active session unless told otherwise, for convenience of use and to minimize the amount of refactoring needed in existing code.  I've deliberately avoided implementing features described in other tickets (like switching the active session) for now, but built this with those functionality expansions in mind so they shouldn't be too tough to add.

### DIFF
--- a/features/test/basic.feature
+++ b/features/test/basic.feature
@@ -73,7 +73,7 @@ Feature: Basic Test Execution
         Scenario: Only run tests with the specified tag
             When a user runs the command "npx specify test --tags '@pass' ./assets/gherkin/binary/"
             Then the command should return a "success" exit code
-        
+
         Scenario: Do not run tests with the specified inverted tag
             When a user runs the command "npx specify test --tags 'not @fail' ./assets/gherkin/binary/"
             Then the command should return a "success" exit code
--- a/eslint.config.ts
+++ b/eslint.config.ts
@@ -98,11 +98,11 @@ export default tseslint.config(
             "**/*.d.ts",
             "**/.DS_Store/**",
             "**/.vscode/**",
+            "**/logs/**",
             "**/node_modules/**",
             "**/dist/**",
             "**/package.json",
             "**/package-lock.json",
-            "logs/",
         ],
     },
 );
--- a/features/test/specify-watch.feature
+++ b/features/test/specify-watch.feature
@@ -1,0 +1,112 @@
+@manual
+Feature: Watch Mode
+    As a software developer
+    I want to run tests in watch mode that automatically rerun when files change
+    So that I can get immediate feedback during development without manually rerunning tests
+
+    Background:
+        Given that the "@specify/core" NPM package is installed
+        And that the "@specify/plugin-cli" NPM package is installed
+        And that a command line prompt is available
+
+    Rule: Watch mode monitors file changes and reruns tests automatically
+
+        @skip
+        Scenario: Watch mode starts and monitors for changes
+            Given that a "passing feature" file exists at "./features"
+            When a user runs the command "npx specify test --watch"
+            Then the command should start in watch mode
+            And the console output should include "Watching for file changes"
+            And the initial test run should complete with a "success" status code
+
+        @skip
+        Scenario: Test reruns automatically when feature file changes
+            Given that a "passing feature" file exists at "./features"
+            And that watch mode is running
+            When the feature file is modified
+            Then the tests should automatically rerun
+
+        @skip
+        Scenario: Test reruns automatically when step definition changes
+            Given that a "passing feature" file exists at "./features"
+            And that a custom step definition file exists at "./features/steps"
+            And that watch mode is running
+            When the step definition file is modified
+            Then the tests should automatically rerun
+
+    Rule: Watch mode can be combined with other options
+
+        @skip
+        Scenario: Watch mode with retry option
+            Given that a "flaky feature" file exists at "./features"
+            When a user runs the command "npx specify test --watch --retry 2"
+            Then the command should start in watch mode
+            And failed tests should be retried according to the retry setting
+
+        @skip
+        Scenario: Watch mode with parallel execution
+            Given that multiple "passing feature" files exist at "./features"
+            When a user runs the command "npx specify test --watch --parallel 2"
+            Then the command should start in watch mode
+            And tests should run in parallel according to the parallel setting
+
+        @skip
+        Scenario: Watch mode with tag filtering
+            Given that a "tagged feature" file exists at "./features"
+            When a user runs the command "npx specify test --watch --tags '@smoke'"
+            Then the command should start in watch mode
+            And only tests matching the tag filter should run
+
+    Rule: Watch mode handles file system events appropriately
+
+        @skip
+        Scenario: Watch mode ignores temporary files
+            Given that a "passing feature" file exists at "./features"
+            And that watch mode is running
+            When a temporary file is created in "./features"
+            Then the tests should not rerun
+
+        @skip
+        Scenario: Watch mode handles file deletion gracefully
+            Given that a "passing feature" file exists at "./features/deleteme.feature"
+            And that watch mode is running
+            When "./features/deleteme.feature" is deleted
+            Then the tests should automatically rerun
+
+        @skip
+        Scenario: Watch mode handles new file creation
+            Given that watch mode is running for "./features"
+            When a new feature file is created at "./features/newtest.feature"
+            Then the tests should automatically rerun
+            And the new test should be included in the test run
+
+    Rule: Watch mode can be stopped gracefully
+
+        @skip
+        Scenario: Watch mode stops on user interrupt
+            Given that watch mode is running
+            When the user sends a SIGINT signal (Ctrl+C)
+            Then the watch mode should stop gracefully
+            And the console output should include "Watch mode stopped"
+
+        @skip
+        Scenario: Watch mode exits with error status when interrupted during test run
+            Given that a "long running feature" file exists at "./features"
+            And that watch mode is running
+            When the user sends a SIGINT signal during test execution
+            Then the watch mode should stop gracefully
+            And the command should exit with an appropriate status code
+
+    Rule: Watch mode option validation
+
+        @skip
+        Scenario: Watch mode conflicts with rerun option
+            When a user runs the command "npx specify test --watch --rerun"
+            Then the command should exit with an "error" status code
+            And the console output should include "conflicting options"
+
+        @skip
+        Scenario: Watch mode requires valid file paths
+            When a user runs the command "npx specify test --watch ./nonexistent/path"
+            Then the command should exit with an "error" status code
+            And the console output should include "path not found error"
--- a/modules/@specify/core/bin/specify.js
+++ b/modules/@specify/core/bin/specify.js
@@ -6,7 +6,18 @@ import process from "node:process";
 import { error } from "node:console";
 
 import("../dist/main.js").catch((err) => {
-    error("Executable hasn't been built yet:", err.message);
+    switch (err.code) {
+        case "ERR_MODULE_NOT_FOUND":
+            error("Module not found. Have you built the project? Run `npm run build`.");
+            break;
+        case "ERR_REQUIRE_ESM":
+            error(
+                "This module requires ESM support. Please ensure your Node.js version supports ESM.",
+            );
+            break;
+        default:
+            error("An unexpected error occurred:", err);
+    }
 
-    process.exit(1);
+    process.exit(2);
 });
--- a/modules/@specify/core/package.json
+++ b/modules/@specify/core/package.json
@@ -25,17 +25,20 @@
   "dependencies": {
     "@cucumber/cucumber": "^11.3.0",
     "@specify/quick-ref": "workspace:*",
+    "chalk": "^5.4.1",
+    "chokidar": "^4.0.3",
+    "deepmerge": "^4.3.1",
     "globby": "^14.1.0",
+    "lodash": "^4.17.21",
     "minimist": "^1.2.8",
-    "serialize-error": "^12.0.0",
-    "type-fest": "^4.41.0"
+    "serialize-error": "^12.0.0"
   },
   "devDependencies": {
     "@types/minimist": "^1.2.5",
     "@types/resolve": "^1.20.6",
-    "deepmerge": "^4.3.1",
     "ts-morph": "^25.0.1",
     "tsc-alias": "^1.8.16",
+    "type-fest": "^4.41.0",
     "typescript": "^5.8.3"
   },
   "engines": {
--- a/modules/@specify/core/src/config/paths.config.ts
+++ b/modules/@specify/core/src/config/paths.config.ts
@@ -1,4 +1,9 @@
-export type PathsConfig = Record<string, string>;
+export type PathsConfig = {
+    [key: string]: string;
+    gherkin: string;
+    logs: string;
+    refs: string;
+};
 
 export const paths: PathsConfig = {
     "gherkin": "./features/**/*.feature",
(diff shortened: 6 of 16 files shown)


modules/@specify/plugin-cli/src/lib/SessionManager.ts
```diff
@@ -0,0 +1,276 @@
...
+     * @param session  - The session to manage
+     * @param name     - The name of the session, for easy reference (optional)
+     * @param activate - Activate the new session
+     */
+    addSession(session: ISystemIOSession, name: string, activate: boolean = true): SessionMeta {
+        const sessionMeta: SessionMeta = { name, session };
+
+        this.#sessions.push(sessionMeta);
+
+        // session.onOutput(this.#processOutput.bind(this));
```
adamlacoste, line 100: I overlooked this line, but it's fixed in the next PR: https://github.com/software-pirates/specify-core/pull/112/files#r2226779787

MentalFusion (changes requested):
A couple of changes requested (marked with checkboxes) and some other misc comments.

modules/@specify/plugin-cli/src/cucumber/support/param_types.ts
```diff
@@ -6,42 +6,44 @@
...
 
+const quotedString = /"(?:\\.|[^\\"])"|'(?:\\.|[^\\'])'/;
+
 defineParameterType({
     "name":   "path",
-    "regexp": /[^"]*/,
+    "regexp": quotedString,
     transformer(input: string): string {
-        return path.resolve(input);
+        return path.resolve(input.slice(1, -1));
```
adamlacoste, line 19: I counted and the function call `stripQuotes()` is 13 characters, exactly the same as `.slice(1, -1)`, so I opted to avoid the question of what to do with that helper function and just used the short-form manipulation directly.

features/test/basic.feature
```diff
@@ -6,99 +6,99 @@ Feature: Basic Test Execution
     Background:
         Given that the "@specify/core" NPM package is installed
         And that the "@specify/plugin-cli" NPM package is installed
-        And that a command line prompt is available
+        And a CLI shell
```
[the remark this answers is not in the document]
wild-love: Im personally in favor of nixing "that" in our step defs, I dont think it really add much to the steps (proper english be damned).

Theres a term for it that I forget, but calling all terminals "shell" is the same as calling all tissues "Kleenex", or all bandages "Bandaids". Id rather we use the more generalized form, such as terminal or just CLI (we could even write it out if yall prefer; command line interface).

modules/@specify/plugin-cli/src/lib/SessionManager.test.ts
```diff
@@ -0,0 +1,254 @@
...
+            expect(sessionManager.activeSession).toBeNull();
+        });
+
+        it("removes one managed session among several", async () => {
+            const altSession = new ShellSession() as unknown as MockShellSession;
+
+            sessionManager.addSession(altSession, "whatever", false);
+
+            expect(sessionManager.sessions.length).toBe(2);
+            expect(sessionManager.activeSession.session).toBe(session);
```
wild-love, line 178: These checks are redundant with above tests that verify the same thing.

modules/@specify/plugin-cli/src/lib/SessionManager.test.ts
```diff
@@ -0,0 +1,254 @@
...
+        });
+
+        it("removes the active session among several", async () => {
+            const altSession = new ShellSession() as unknown as MockShellSession;
+
+            sessionManager.addSession(altSession);
+
+            expect(sessionManager.sessions.length).toBe(2);
+            expect(sessionManager.activeSession.session).toBe(altSession);
```
wild-love, line 192: We could get rid of these two checks, since its tested above.

modules/@specify/plugin-cli/src/lib/SessionManager.test.ts
```diff
@@ -0,0 +1,254 @@
...
+
+            sessionManager.removeSession();
+
+            expect(sessionManager.sessions.length).toBe(1);
```
wild-love, line 196: What do you think about changing 
`expect(sessionManager.sessions.length).toBe(1);`
to
`expect(sessionManager.sessions).toBeEqual([session]);`? 
Seems a bit more precise to me.

modules/@specify/plugin-cli/src/lib/SessionManager.test.ts
```diff
@@ -0,0 +1,254 @@
...
+            expect(sessionManager.activeSession.session).toBe(session);
+        });
+    });
+
+    describe("run()", () => {
+        it("writes command to session and resolves after delimiter", async () => {
```
wild-love, line 202: We should remove the delimiter check from this test and focus solely on checking the writing. The delimiter resolution is tested immediately below.

modules/@specify/plugin-cli/src/lib/SessionManager.ts
```diff
@@ -0,0 +1,276 @@
+/**
+ * SessionManager Module
+ *
+ * Provides an interface for executing interactive shell commands via
+ * ISystemIOSession. Commands are tracked with delimiters and expose
+ * output and status code results.
```
wild-love, line 6: "exit" code results 😡

modules/@specify/plugin-cli/src/lib/SessionManager.ts
```diff
@@ -0,0 +1,276 @@
...
+ */
+interface SessionMeta {
+    commandResolve?: () => void;
+    curCommand?: string;
+    delimiter?: Delimiter;
+    exitCode?: number;
+    name?: string;
+    output?: string;
+    session: ISystemIOSession;
+}
```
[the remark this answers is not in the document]
wild-love: Im actually a bit uneasy about prefixing `I` to interfaces that arent _ackshually_ interfaces. I named it `ISystemIOSession` because it was meant to be implemented as an interface by a class. `IDelimiter` and co. pollutes that notation since they are merely types in practice.

modules/@specify/plugin-cli/src/lib/SessionManager.ts
```diff
@@ -0,0 +1,276 @@
...
+     */
+    removeSession(sessionMeta: ISessionMeta): void {
+        sessionMeta ??= this.#activeSession;
+
+        const index = this.#sessions.indexOf(sessionMeta);
+
+        this.#sessions.splice(index, 1);
+
+        if (this.#activeSession === sessionMeta) {
+            const prevIndex = index === 0 ? this.#sessions.length - 1 : index - 1;
```
wild-love, line 161: Instead of this, I propose we keep track of the previous active session and swap back to that when the current active session is removed. In this way, it would behave like OS windows, which I find to be pretty intuitive.

If we do that, we have to remember to do `prevActiveSession = null`. This only works with 2 sessions, however. With more, we'd have to build an array that is re-ordered according to usage. I dont think it'd be complex, but perhaps out of scope for this ticket.

modules/@specify/plugin-cli/src/lib/SessionManager.ts
```diff
@@ -0,0 +1,276 @@
...
+            // if so, unset active session
+            // if not, set active session to the previous index
+            this.#activeSession = prevIndex < 0 ? null : this.#sessions[prevIndex];
+        }
+    }
+
+    /**
+     * Executes a single command within a managed session.
+     *
+     * Only one command may be activeat a time in any given session. Output and
```
wild-love, line 173: "active at"

modules/@specify/plugin-cli/src/lib/SessionManager.ts
```diff
@@ -0,0 +1,276 @@
...
+    /**
+     * Handles raw output received from the session.
+     *
+     * If the delimiter is found in the output, the command is considered
+     * complete and its status code is recorded.
+     *
+     * @param output      - The unmodified session output
+     * @param sessionMeta - The session to process output for (optional)
+     */
+    #processOutput(output: string, sessionMeta: ISessionMeta): void {
```
wild-love, line 246: `sessionMeta` is required here, but below we check if its set anyway. So should this be an optional argument?

modules/@specify/plugin-cli/src/lib/SessionManager.ts
```diff
@@ -0,0 +1,276 @@
...
+            this.#resolveRun(sessionMeta);
+        }
+    }
+
+    /**
+     * Resolves the running command's promise.
+     *
+     * @param sessionMeta - The session for the running command (optional)
+     */
+    #resolveRun(sessionMeta: ISessionMeta): void {
```
wild-love, line 267: Same minor issue here as in `#processOutput`

wild-love (changes requested):
Changes, please!

modules/@specify/plugin-cli/src/lib/SessionManager.test.ts
adamlacoste: That comparison would fail, as `sessions` is an array of `SessionMeta` objects.

modules/@specify/plugin-cli/src/lib/SessionManager.test.ts
adamlacoste: I don't see where we are checking the delimiter here.  But the structure of this test is a carry-over from your previous `run()` unit tests, just modified to work following the async conversion.  I tried to change as little as possible to get it working again.

modules/@specify/plugin-cli/src/lib/SessionManager.ts
adamlacoste: Whoops, good catch.

- [ ] Search for "status" in the CLI plugin to catch this and any other stragglers.

modules/@specify/plugin-cli/src/lib/SessionManager.ts
adamlacoste: Yeah, that seems out of scope to me.

modules/@specify/plugin-cli/src/lib/SessionManager.ts
adamlacoste: - [ ] Fix typo in `run` docblock

modules/@specify/plugin-cli/src/lib/SessionManager.ts
adamlacoste: It is optional; if this method is called without a second param it will be `undefined` until being overwritten with the first line below.

modules/@specify/plugin-cli/src/lib/SessionManager.test.ts
```diff
@@ -0,0 +1,254 @@
...
+    });
+
+    describe("kill()", () => {
+        it("resolves kill() on session close", async () => {
+            let resolved = false;
+
+            const promise = sessionManager.kill().then(() => (resolved = true));
+
+            // wait arbitrary time to ensure promise hasn't resolved
+            await new Promise((resolve) => setTimeout(resolve, 100));
```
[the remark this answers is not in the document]
MentalFusion: I did attempt to add this sort of thing to my `TestCommandWatcher` test, and it caused all sorts of chaos and things broke in a variety of ways. I'm in full support of Wild tackling it in a separate ticket.

modules/@specify/plugin-cli/src/lib/SessionManager.ts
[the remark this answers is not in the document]
MentalFusion: Well that's just silly.

modules/@specify/plugin-cli/src/lib/SessionManager.ts
MentalFusion: Since `sessionMeta` is optional, we should define it as such in the function definition:

```ts
    // add ? to optional param; TSDoc doesn't need (optional) annotation (but we can keep it)
    #processOutput(output: string, sessionMeta?: ISessionMeta): void {
        ...
    }
```

modules/@specify/plugin-cli/src/lib/SessionManager.ts
MentalFusion: Same issue from me as well: optional params should have `?` to denote their optional state. The TSDoc doesn't need to be explicitly annotated with `(optional)`, but we can definitely keep the convention if we want (I haven't been using it myself so far since the `?` makes it apparent the param is optional without the explicit "(optional)" label).

MentalFusion (changes requested):
Mostly just revisit the params that are optional and add `?` where they're optional.

modules/@specify/plugin-cli/src/lib/SessionManager.ts
adamlacoste: I didn't know this was a thing; I will go back and mark optional params as such, and remove the TSDoc notation since that's redundant.

- [ ] Mark optional params with `?` throughout `SessionManager`

modules/@specify/plugin-cli/src/lib/SessionManager.test.ts
wild-love: I think its fine to trust in the tests elsewhere, and doing it this way means that if the implementation changes for the redundant checks' underlying code that breaks the tests, we're going to have to fix it in more than one test. Put another way, I think this violates DRY.

If y'all are intent on doing it this way, then I'll accept that I'm outvoted. But for posterity I think it's better to rely on the previous tests.

modules/@specify/plugin-cli/src/lib/SessionManager.test.ts
wild-love: Then it was me who was bad and I should feel bad 😡 

The lines Im referring to are:
```
session.emitDelimiter(0);
await promise;
```
Neither of these add anything to the test. And if its changed, the test should be renamed "writes command to session ".

Could also just leave it for another time, its a bit out of scope.

modules/@specify/plugin-cli/src/lib/SessionManager.ts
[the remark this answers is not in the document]
wild-love: If the case comes where we want `ISessionMeta` to be extended as an interface, we could always just change it at that point. Id prefer for the code to represent how things are now or how we plan for them to be in the future, not how they could possibly be someday.

All interfaces prefixed by `I` is fine by me. Lets just be conscious of not using interfaces as type, in that case..

wild-love (changes requested):
Further discussion.

modules/@specify/plugin-cli/src/lib/SessionManager.ts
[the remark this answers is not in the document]
adamlacoste: So to state this up front, I don't really have a dog in this interface vs. type fight, but I didn't understand why we were changing from one to the other, so I decided to do some reading into the practical differences and what guidelines I should be using in choosing one or the other.  I found a lot of opinions online, many of them contradictory, but finally one of them linked me to a place in the official documentation which addresses this issue:

https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces

I'll quote from the end of this section:

> For the most part, you can choose based on personal preference, and TypeScript will tell you if it needs something to be the other kind of declaration. **If you would like a heuristic, use interface until you need to use features from type.**

(emphasis mine)

So that's pretty official, right?  It seems that these should remain interfaces unless there is a good reason to make them types instead.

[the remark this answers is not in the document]
adamlacoste: Per our discussion this morning, we will leave these interfaces alone and consider revising the naming conventions thereof in a different ticket.

adamlacoste:
All feedback resolved.

MentalFusion (approved):

PR #113, opened 2025-07-29
Cucumber extraction :cucumber: 

wild-love:


modules/@specify/core/src/lib/Logger.integration.test.ts
wild-love, line 1: What do we think about this as a convention, where we use `.integration.test.ts` for integration tests? In theory, it should make it easy to include/exclude/target integration tests in the future.

modules/@specify/core/src/lib/Logger.ts
wild-love, line 1: Extraction of this class was useful since `CucumberTool` shouldnt care about system IO.

modules/@specify/core/src/lib/TestCommand.test.ts
wild-love, line 1: This file and `TestCommand.ts` changed a lot due to the extraction of all cucumber stuff. The benefit? No more Cucumber output in the unit tests!

modules/@specify/core/src/lib/Logger.test.ts
```diff
@@ -0,0 +1,145 @@
...
+            vi.spyOn(crypto, "randomUUID").mockReturnValue(fakeUUID);
+
+            const logger = new Logger();
+            const prefix = "test-prefix";
+            const path   = await logger.generateTmpLogPath(prefix);
+
+            expect(path).toBe(expectedPath(prefix));
+        });
+
+        it("creates the specify subdirectory if it doesnt already exist", async () => {
```
MentalFusion, line 87: Missing apostrophe in `doesn't`.

- [ ] Fix spelling

MentalFusion (changes requested):

modules/@specify/core/src/lib/Logger.integration.test.ts
wild-love: Naming it `Logger.integration.ts` isnt initially clear that its a test, and if its not named `...test.ts` then we wont get the handy file name coloring thats usually associated with test files.

We should be able to just modify our `package.json` script with something like `vitest run "**/*.integration.*"` and  `vitest run "**/*^(integration).test.ts"` (or something similar). So we could do `pnpm test:unit` or `pnpm test:integration`.

modules/@specify/core/src/lib/Logger.integration.test.ts
```diff
@@ -0,0 +1,74 @@
+import { existsSync    } from "node:fs";
+import { rm, writeFile } from "node:fs/promises";
```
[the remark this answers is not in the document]
wild-love: - [ ] convert to sync forms

modules/@specify/core/src/lib/Logger.test.ts
```diff
@@ -0,0 +1,145 @@
+import os         from "node:os";
+import { Logger } from "./Logger";
+
+const testFileFactory = (filename: string) => {
+    const json = { filename };
+    const path = `/${filename}.json`;
+
+    return {
+        [path]:{
```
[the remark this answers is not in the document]
MentalFusion: Appears to be a bug with our lint rule when using variable substitution for property names.

Issue #115, opened 2025-07-30
Rename all interfaces to remove the I-prefix


adamlacoste:
Previously we had been using a naming convention for interfaces which prefixed them with the letter `I`, but the team has decided to reverse this policy and remove all I-prefixes from interfaces.

PR #116, opened 2025-07-30
[#115] Renamed any interface defined with an I-prefix and updated the…

MentalFusion:
…ir references as well to remove the I-prefix.

### DIFF
--- a/modules/@specify/quick-ref/src/lib/QuickRef.test.ts
+++ b/modules/@specify/quick-ref/src/lib/QuickRef.test.ts
@@ -40,6 +40,7 @@ describe("QuickRef", () => {
     it("allows read-only access to refs", () => {
         const quickRef = new QuickRef();
 
+        // @ts-expect-error - Testing read-only access
         expect(() => (quickRef.refs = {})).toThrow();
     });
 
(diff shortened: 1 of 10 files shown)


modules/@specify/quick-ref/src/lib/QuickRef.test.ts
```diff
@@ -40,6 +40,7 @@ describe("QuickRef", () => {
     it("allows read-only access to refs", () => {
         const quickRef = new QuickRef();
 
+        // @ts-expect-error - Testing read-only access
         expect(() => (quickRef.refs = {})).toThrow();
```
MentalFusion, line 44: The comment above this line silences TypeScript's type-check error as we are expecting the error when testing this read-only property.

adamlacoste: Do we need to do this everywhere we are testing throws?

MentalFusion: No. We had to do it here because we're testing whether or not assigning to a read-only property will throw, and TypeScript already throws / errs when attempting that and builds will break.

Technically, this test isn't needed since TypeScript has protections built in for this action, but if we want to be thorough and test it ourselves, we need to tell TypeScript we expect this to error since we're performing a unit test.

wild-love: I lean more towards avoiding these kinds of tests that essentially just test that typescript is doing its job. Wont hurt anything to keeps it, but its a bit pointless 🤷

PR #119, opened 2025-08-07
Exit codes

adamlacoste:


features/plugin-cli/steps/shell.feature
```diff
@@ -33,7 +33,7 @@ Feature: Shell Step Definitions
             Given a CLI shell
             When the user starts the async command "for num in $(seq 1 3); do sleep 1; echo $num; done"
             And the user waits for the last command to return
-            Then the command should return a 0 exit code
+            Then the last command's exit code should be a 0
```
adamlacoste, line 36: As discussed on our call today, I would have preferred to make the article which precedes the integer at the end (`a/an`) optional, but Cucumber apparently does not allow alternation within an optional clause.  My short-term fix was to make the article `a/an` mandatory, even though it's not really an ideal solution.  I'm open to exploring other ways of phrasing this step to be a bit more naturalistic; just bear in mind that we want to make sure we are considering both literal value (integer, in this case) and ref name variants.

For your consideration, here are some suggestions for alternative phrasing which avoid the optional/alternate problem:

```
Then the last command's exit code should be 0
Then the last command's exit code should be $success
```

```
Then the last command should return an exit code of 0
Then the last command should return an exit code of $success
```

```
Then the last command should return a/an 0 exit code
Then the last command should return a/an $success exit code
```

modules/@specify/plugin-cli/src/cucumber/step_definitions/shell.ts
```diff
@@ -132,7 +132,7 @@ function verifyExitCode(exitCode: number): void {
     assert.equal(
         this.cli.manager.exitCode,
         exitCode,
-        "The command's status code did not match expectations.",
+        "The command's exit code did not match expectations.",
```
adamlacoste, line 135: We weren't being consistent here.  We support both `exit code` and `exit status` phrasings but we've been using `exit code` everywhere else throughout our code.

features/plugin-cli/steps/shell.feature
wild-love: As an alternate alternative, Im in favor of the solution you mentioned in the team chat; just defining multiple expressions that match what we want. Its not an ideal solution, but I dont see this being a problem often, and even if it is I dont think its a big deal as long as we keep the "aliases" grouped.

features/test/retry.feature
```diff
@@ -20,13 +20,14 @@ Feature: Retry Flaky Tests
         Scenario: A test run that fails with 2 retries
             Given that a "3rd retry passing feature" file exists at "./features"
             When a user runs the command "npx specify test --retry 2"
-            Then the command should return a $failure exit code
+            # Then the command should return a $failure exit code
+            Then the last command's exit code should be a $success
```
wild-love, line 24: Why did this change? Seems like this is not the result we want for this test.

adamlacoste: That's a mistake, thanks for catching that

- [ ] Fix the mistake that changed failure to success

MentalFusion (approved):

Issue #80, opened 2025-07-15
Step Def: Switch active shell


adamlacoste:
Following the completion of #79, we need to add a new step definition that cycles between active shell instances:

* `When a/the user switches shells`

PR #122, opened 2025-08-12
[#80] Added the ability for a user to switch active shells. Shells ar…

MentalFusion:
…e swapped in the order they were opened starting from the currently active shell.


adamlacoste (approved):

modules/@specify/plugin-cli/src/lib/SessionManager.test.ts
```diff
@@ -357,6 +357,35 @@ describe("SessionManager", () => {
...
+    describe("switchToNextSession()", () => {
+        let altSession: MockShellSession;
+
+        beforeEach(() => {
+            altSession = new ShellSession() as unknown as MockShellSession;
+
+            sessionManager.addSession(session);
+            sessionManager.addSession(altSession, "altSession", false);
+
+            expect(sessionManager.activeSession.session).toBe(session);
```
wild-love, line 369: Its ok to make this check if you want, but I don't think we should be testing things in the test setup.

- [ ] Move assertion out of `beforeEach()`

modules/@specify/plugin-cli/src/lib/SessionManager.test.ts
```diff
@@ -357,6 +357,35 @@ describe("SessionManager", () => {
...
+        });
+
+        it("switches to the next managed session", () => {
+            sessionManager.switchToNextSession();
+
+            expect(sessionManager.activeSession.session).toBe(altSession);
+        });
+
+        it("switches to the first managed session if already at the last one", () => {
```
wild-love, line 378: I think we should add a third session here and make sure we really are going from 3rd to 1st, instead of just swapping between 2nd and 1st (regardless of how many sessions we might have). We dont have to though, its up to you.

modules/@specify/plugin-cli/src/lib/SessionManager.ts
```diff
@@ -257,6 +257,19 @@ export class SessionManager {
         return newCmdMeta;
     }
 
+    /**
+     * Switch to the next session in the list.
+     *
+     * @remarks
+     * If there is no next session, the first session in the list is activated.
+     */
```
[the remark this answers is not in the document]
wild-love: This is a 10/10 solution :fire: :fire: :fire:

wild-love (changes requested):

modules/@specify/plugin-cli/src/lib/SessionManager.test.ts
wild-love: Im not so sure. I think there is a difference between swapping between `primarySession` and `secondarySession` vs looping through an aribrary number of sessions in order.

The current test would pass regardless if we were swapping vs looping, whereas adding a 3rd session would ensure we were looping and not swapping. We know the current implementation is correct, but this test doesnt fully validate the intended behavior.

features/plugin-cli/steps/shell.feature
```diff
@@ -89,7 +89,19 @@ Feature: Shell Step Definitions
...
+            And that the "@specify/plugin-cli" NPM package is installed
+            And a CLI shell
+            And another CLI shell
+
+        Scenario: Swap between shells in sequence
+            When a user runs the command "echo 'First shell!'"
+            And a user switches shells
+            And a user runs the command "echo 'Second shell!'"
+            Then the last command's terminal output should match "Second shell!"
+            When a user switches shells
```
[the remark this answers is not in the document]
adamlacoste: I disagree with this.  This test _is_ testing one thing, but it's validating it at multiple points in the process, which is entirely appropriate IMO.  Splitting the test would not make it less brittle, it would just make it take longer.

wild-love (approved):

Issue #146, opened 2025-08-22
Bug: User override config is covering execution bugs following base config trimming


adamlacoste:
When I run Cucumber from a different directory than the repo root -- meaning that the specify.config.json file serving as the project's user override config is not detected -- I get an error that I don't see normally, which is caused by attempting to add to a config `import` array that doesn't necessarily exist.

```
An unexpected error occurred: TypeError: Cannot read properties of undefined (reading 'push')
    at file:///Users/adam/Development/specify/core/modules/@specify-bdd/specify/dist/main.js:29:20
```

We need to fix this and any other bugs which exist now that we don't initialize these properties in the base cucumber config, but were being covered by a user override which did.

PR #148, opened 2025-08-25
Cucumber config bugs

adamlacoste:
[#146] Fixed a couple bugs that were being covered by our project's user override config

### DIFF
--- a/modules/@specify-bdd/specify/src/config/debug.config.ts
+++ b/modules/@specify-bdd/specify/src/config/debug.config.ts
@@ -1,3 +1,3 @@
 export type DebugConfig = boolean;
 
-export const debug: DebugConfig = true;
+export const debug: DebugConfig = false;
--- a/modules/@specify-bdd/specify/src/main.ts
+++ b/modules/@specify-bdd/specify/src/main.ts
@@ -34,6 +34,7 @@ for (const plugin of config.plugins) {
 }
 
 // add core cucumber support code path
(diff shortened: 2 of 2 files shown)


modules/@specify-bdd/specify/src/config/debug.config.ts
```diff
@@ -1,3 +1,3 @@
 export type DebugConfig = boolean;
 
-export const debug: DebugConfig = true;
+export const debug: DebugConfig = false;
```
adamlacoste, line 3: This might not be a "bug" per se but it was definitely not intended that debug mode should be enabled by default, as it makes Cucumber output extraordinarily verbose.

modules/@specify-bdd/specify/src/main.ts
```diff
@@ -34,6 +34,7 @@ for (const plugin of config.plugins) {
 }
 
 // add core cucumber support code path
+cucumberCfg.import ??= [];
 cucumberCfg.import.push(path.resolve(import.meta.dirname, "cucumber"));
```
adamlacoste, line 38: This was the actual bug I had previously encountered; because we no longer set a default for `cucumber.import` in the base config, this `Array.push` will throw an error, but only if you don't have a user override config assigning that key a value.  Our project user override config does, so the bug was buried until I tried testing it without that.

wild-love (approved):

Issue #151, opened 2025-08-27
Prevent Specify test config pollution


adamlacoste:
At present, when we use Specify to run tests which themselves run Specify, the same project-level override config applies to both Specify executions.  This can be a problem because some tests of specify assume certain default behaviors which might be changed by that override config.  We must therefore ensure that the tests which run Specify do so in an isolated context, free from the influence of the project's override config.

PR #152, opened 2025-08-27
Only YOU can prevent config pollution

adamlacoste:
[#151] added new step defs and revised existing gherkin specs to protect against config pollution

### DIFF
--- a/features/test/basic.feature
+++ b/features/test/basic.feature
@@ -7,21 +7,22 @@ Feature: Basic Test Execution
         Given that the "@specify-bdd/specify" NPM package is installed
         And that the "@specify-bdd/plugin-cli" NPM package is installed
         And a CLI shell
+        And that the working directory is "./assets/gherkin"
 
     Rule: The run should succeed only if all tests pass
 
         Scenario: All tests pass
-            When a user runs the command "npx specify test ./assets/gherkin/binary/passing.feature"
+            When a user runs the command "npx specify test ./binary/passing.feature"
             Then the last command's exit code should be a $success
             And the last command's terminal output should match $passingTestResult
 
         Scenario: All tests fail
-            When a user runs the command "npx specify test ./assets/gherkin/binary/failing.feature"
+            When a user runs the command "npx specify test ./binary/failing.feature"
             Then the last command's exit code should be a $failure
             And the last command's terminal output should match $failingTestResult
 
         Scenario: Mixed pass/fail tests
-            When a user runs the command "npx specify test ./assets/gherkin/binary/"
+            When a user runs the command "npx specify test ./binary/"
             Then the last command's exit code should be a $failure
 
     Rule: The run should allow verification of string values in terminal output
@@ -34,12 +35,12 @@ Feature: Basic Test Execution
 
         @dependency
         Scenario: Feature has a Gherkin syntax error
-            When a user runs the command "npx specify test ./assets/gherkin/invalid.feature"
+            When a user runs the command "npx specify test ./invalid.feature"
             Then the last command's exit code should be an $error
 
         @dependency
         Scenario: Feature contains undefined step definitions
-            When a user runs the command "npx specify test ./assets/gherkin/undefined.feature"
+            When a user runs the command "npx specify test ./undefined.feature"
             Then the last command's exit code should be an $error
 
     Rule: The run should error if there are no available tests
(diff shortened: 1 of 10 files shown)


features/plugin-cli/steps/file_system.feature
adamlacoste, line 1: This solution required the addition of a new, general-purpose step def which changes the working directory of the active shell.  I went ahead and started a new feature file exercising it.

features/test/basic.feature
```diff
@@ -7,21 +7,22 @@ Feature: Basic Test Execution
         Given that the "@specify-bdd/specify" NPM package is installed
         And that the "@specify-bdd/plugin-cli" NPM package is installed
         And a CLI shell
+        And that the working directory is "./assets/gherkin"
```
adamlacoste, line 10: Every feature file that needs to avoid config pollution (which ought to be all of them) can add a step like this in the `Background:` section; the current config logic only looks in the current working directory for a `specify.config.json` file.

features/test/basic.feature
```diff
@@ -7,21 +7,22 @@ Feature: Basic Test Execution
...
 
     Rule: The run should succeed only if all tests pass
 
         Scenario: All tests pass
-            When a user runs the command "npx specify test ./assets/gherkin/binary/passing.feature"
+            When a user runs the command "npx specify test ./binary/passing.feature"
```
adamlacoste, line 15: Because the working directory changed, all relative paths had to be updated.

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/file_system.ts
adamlacoste, line 1: This file contained a bunch of old code for copying files around that was built around an approach we're not using anymore.  While some of it might need to be rebuilt later, it will undoubtably be in a much different form, so I didn't see much point in keeping the commented-out legacy stuff in here.

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/file_system.ts
```diff
@@ -4,48 +4,34 @@
...
  */
-// async function writeFile(file: FileParam, destPath: string): Promise<void> {
-//     const opts = {
-//         "encoding": file.encoding || "utf8",
-//         "mode": file.mode || 0o644,
-//     } as Partial<ObjectEncodingOptions>;
+async function changeDirectory(dirPath: string): Promise<void> {
+    assert.ok(this.cli.manager, new AssertionError({ "message": "No shell session initialized." }));
+
+    this.cli.manager.run(`cd ${dirPath}`); // TODO: generalize this so it will work for any OS/shell
```
adamlacoste, line 28: I'm not sure at this time the best way to make this step def OS/shell agnostic, but we'll need to figure it out sooner or later.

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
```diff
@@ -97,11 +97,11 @@ Then(
  *
  * @param command - The command to run
  *
- * @throws {@link AssertionError}
+ * @throws AssertionError
```
adamlacoste, line 100: I fixed all these link errors Michael alerted us to while I was in here.

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
```diff
@@ -97,11 +97,11 @@ Then(
...
  * If there is no SessionManager initialized.
  */
 function execCommand(command: string): void {
-    assert.ok(this.cli.manager);
+    assert.ok(this.cli.manager, new AssertionError({ "message": "No shell session initialized." }));
```
adamlacoste, line 104: While I was at this, I went ahead and normalized the way we're throwing errors with our `assert` statements.  Now all step defs which make an assertion will do so in a consistent way and with a custom error message.

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
```diff
@@ -184,32 +184,40 @@ function verifyMatchingOutput(pattern: RegExp | string): void {
 /**
  * Verify that the last command's execution time is the specified number of seconds or less.
  *
- * @param seconds - The maximum number of seconds that should have elapsed
+ * @param timeout - The maximum number of seconds that should have elapsed
```
adamlacoste, line 187: I should have brought this up in code review, but somehow it escaped my notice.  If you disagree with this change, I'll revert it.

While I was in here messing with the assertion error messages, I noticed that this one was made very long (breaking our line limit) by the math being done in the middle of it.  I decided to break that out into a constant, which I named `elapsed`.  Having done so, however, I thought that the param name `seconds` was now kind of ambiguous because `elapsed` was also a number of seconds, so I renamed that as well.

specify.config.json
```diff
@@ -1,6 +1,7 @@
 {
     "cucumber": {
-        "format": [ "progress" ]
+        "format": [ "progress" ],
+        "parallel": 3
```
adamlacoste, line 4: Restored to its rightful place.

PR #153, opened 2025-08-29
Parallel option constraints

wild-love:


features/test/parallel.feature
```diff
@@ -25,20 +25,20 @@ Feature: Parallel Execution
...
-        @skip @todo
-        Scenario: A value of 0 is equivalent to no limit
+        Scenario: A value of 0 is rejected
+            When a user runs the command "npx specify test --parallel 0"
+            Then the last command's exit code should be an $failure
+            And the last command's terminal output should match $invalidParallelError
 
-        @skip @todo
+        @skip @todo @review
         Scenario: An excessively high value is rejected (or handled gracefully)
```
wild-love, line 44: Time for a review of this scenario :monocle_face:

adamlacoste: Conclusion from call: just remove this scenario, we don't need to set a limit.

features/test/parallel.feature
```diff
@@ -25,20 +25,20 @@ Feature: Parallel Execution
...
-            And the console output should match "help message"
+            Then the last command's exit code should be an $failure
+            And the last command's terminal output should match $invalidParallelError
 
-        @skip
         Scenario: Non-numeric values are rejected
             When a user runs the command "npx specify test --parallel 'bad-value'"
-            Then the last command's exit code should be an $error
-            And the console output should match "help message"
+            Then the last command's exit code should be an $failure
```
MentalFusion, line 35: Same thing here. Anywhere it reads `an $failure` should be updated to `a $failure` for better reading.

modules/@specify-bdd/specify/src/lib/cli-parsers.test.ts
```diff
@@ -0,0 +1,42 @@
...
+    });
+
+    it("throws for zero", () => {
+        expect(() => parseParallelOption("0")).toThrow(error);
+    });
+
+    it("throws for negative numbers", () => {
+        const number = Math.floor(Math.random() * -1000);
+
+        expect(() => parseParallelOption(number.toString())).toThrow(error);
```
MentalFusion, line 22: The random number generation makes this test non-deterministic. We should just use `-1` or some other fixed negative number.

modules/@specify-bdd/specify/src/lib/cli-parsers.test.ts
```diff
@@ -0,0 +1,42 @@
...
+    });
+
+    it("throws for non-integer numbers", () => {
+        expect(() => parseParallelOption(Math.random().toString())).toThrow(error);
```
MentalFusion, line 26: Same thing as above; we should just test a fixed float value (e.g., `1.2` or something).

MentalFusion (changes requested):

modules/@specify-bdd/specify/src/lib/cli-parsers.test.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    it("throws for non-numeric strings", () => {
+        expect(() => parseParallelOption("abc")).toThrow(error);
+        expect(() => parseParallelOption("")).toThrow(error);
+    });
+
+    it("throws for null and undefined", () => {
+        expect(() => parseParallelOption(null)).toThrow(error);
+        expect(() => parseParallelOption(undefined)).toThrow(error);
+    });
```
MentalFusion, line 37: Now that I think about it, we could just combine these invalid input tests:

```ts
it("throws for invalid inputs", () => {
    expect(() => parseParallelOption("0")).toThrow(error);
    expect(() => parseParallelOption("-1")).toThrow(error);
    expect(() => parseParallelOption("1.5")).toThrow(error);
    expect(() => parseParallelOption("abc")).toThrow(error);
    expect(() => parseParallelOption("")).toThrow(error);
});
```

There's not a lot of reason to separating them out beyond adding extra numbers to our passing test count (which isn't nothing, I guess).

modules/@specify-bdd/specify/src/lib/cli-parsers.test.ts
adamlacoste: It should work regardless of what number we generate, but I agree that it is better to just use `-1`.  If we think there is a realistic chance that different negative numbers could produce different results, we should test it with multiple, explicitly defined negative numbers.  Any bug exposed by a randomly generated number might not be revealed if the wrong number is generated, after all.  We want the tests to produce consistent results.

adamlacoste (changes requested):

adamlacoste (dismissed):
I still want to change the behavior of Commander to return 2 rather than 1, but that's scope creep and can be a separate ticket.

modules/@specify-bdd/specify/src/lib/cli-parsers.test.ts
MentalFusion: To be fair, you'll still get the exact location of what failed in the test.

MentalFusion (approved):
Looks good!

Issue #133, opened 2025-08-18
Feature: Rerun failed tests


adamlacoste:
Implement the ability to rerun tests which failed in the previous execution, as documented in runner/features/rerun.feature.

PR #169, opened 2025-09-09
Rerun spec revised

adamlacoste:
[#133] Revised rerun feature spec

### DIFF
--- a/features/test/basic.feature
+++ b/features/test/basic.feature
@@ -90,10 +90,10 @@ Feature: Basic Test Execution
 
         Scenario: Unsupported option
             When a user runs the command "npx specify --bad-option"
-            Then the last command's exit code should be a ${ref.exitCode.failure}
+            Then the last command's exit code should be a ${ref.exitCode.error}
             And the last command's terminal output should match ${ref.terminalOutput.invalidOptionMessage}
 
         Scenario: Mix of supported and unsupported options
             When a user runs the command "npx specify --tags '@pass' --bad-option"
-            Then the last command's exit code should be a ${ref.exitCode.failure}
+            Then the last command's exit code should be a ${ref.exitCode.error}
             And the last command's terminal output should match ${ref.terminalOutput.invalidOptionMessage}
--- a/features/test/parallel.feature
+++ b/features/test/parallel.feature
@@ -27,15 +27,15 @@ Feature: Parallel Execution
 
         Scenario: Floats are rejected
             When a user runs the command "npx specify test --parallel 0.5"
-            Then the last command's exit code should be a ${ref.exitCode.failure}
+            Then the last command's exit code should be a ${ref.exitCode.error}
             And the last command's terminal output should match ${ref.terminalOutput.invalidParallelError}
 
         Scenario: Non-numeric values are rejected
             When a user runs the command "npx specify test --parallel 'bad-value'"
-            Then the last command's exit code should be a ${ref.exitCode.failure}
+            Then the last command's exit code should be a ${ref.exitCode.error}
             And the last command's terminal output should match ${ref.terminalOutput.invalidParallelError}
 
         Scenario: A value of 0 is rejected
             When a user runs the command "npx specify test --parallel 0"
-            Then the last command's exit code should be a ${ref.exitCode.failure}
+            Then the last command's exit code should be a ${ref.exitCode.error}
             And the last command's terminal output should match ${ref.terminalOutput.invalidParallelError}
--- a/features/test/retry.feature
+++ b/features/test/retry.feature
@@ -67,12 +67,12 @@ Feature: Retry Flaky Tests
 
         Scenario: Floats are rejected
             When a user runs the command "npx specify test --retry 0.5 ./binary/passing.feature"
-            Then the last command's exit code should be a ${ref.exitCode.failure}
+            Then the last command's exit code should be a ${ref.exitCode.error}
             And the last command's terminal output should match ${ref.terminalOutput.invalidRetryError}
 
         Scenario: Non-numeric values are rejected
             When a user runs the command "npx specify test --retry 'bad-value' ./binary/passing.feature"
-            Then the last command's exit code should be a ${ref.exitCode.failure}
+            Then the last command's exit code should be a ${ref.exitCode.error}
             And the last command's terminal output should match ${ref.terminalOutput.invalidRetryError}
 
     # Rule: Retry tag option only accepts well-formed tag expressions


features/test/rerun.feature
```diff
@@ -9,38 +9,71 @@ Feature: Rerun Failed Tests
...
-            And the code under test is modified so the issue is resolved
-            And the user runs the command "npx specify test --rerun"
-            Then the last command's exit code should be a $success
-            And the console output should match $passingTestResult
-
-        @skip
-        Scenario: Rerun is attempted with no prior test run
-            Given that a "failing feature" file exists at "./features"
-            When the user runs the command "npx specify test --rerun"
+            And the last command's terminal output should match $noTestCasesError
```
wild-love, line 61: We should reconsider this. I'm too involved in the project to be able to say for sure, but I think my intuition would be that no tests run via `rerun` would just be a no-op, not an error.

Running `test` and having zero tests is for sure an error, but Im not so sure about `rerun`.

wild-love (changes requested):
I think we need a different solution for ensuring the rerun file doesnt exist. I left a possible solution in a diff comment.

features/test/basic.feature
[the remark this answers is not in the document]
adamlacoste: Although it was not the primary reason for the change, this new approach removes the implicit type-matching we were doing between step def param fields and refs.  Previously, the step def param type would be used as the first level key in the ref lookup; for instance, a `{ref:exitCode}` param named `$success` would look in the refs file under `exitCode.success`.  Now the key chain is explicitly included in the ref notation, which means that users are free to structure the refs file however they want.  This may mean we need to do some more defensive coding in the step defs as a result, since the input can be anything.  (Although really, that was a problem before too, since the refs JSON had no type enforcement, only classification.)

features/test/rerun.feature
```diff
@@ -7,40 +7,75 @@ Feature: Rerun Failed Tests
         Given that the "@specify-bdd/specify" NPM package is installed
         And that the "@specify-bdd/plugin-cli" NPM package is installed
         And a CLI shell
+        And a temp file path referenced as "file.rerun.path" # generates a unique temp file path and stores it as ref
```
adamlacoste, line 10: The inline comment here should be removed once this is implemented; I just left it as a reminder for my intentions with this step.

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/file_system.ts
```diff
@@ -7,9 +7,9 @@
 import { Given, When            } from "@cucumber/cucumber";
 import assert, { AssertionError } from "node:assert/strict";
 
-Given("(that )the working directory is {path}", changeDirectory);
+Given("(that )the working directory is {filePath}", changeDirectory);
```
adamlacoste, line 10: Changed this because `path` is so vague and can refer to a lot of things (a relative URL, for instance) but the param type definition was explicitly using `path.resolve` to force this into an absolute file path.

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/packages.ts
```diff
@@ -7,7 +7,7 @@ import { Given   } from "@cucumber/cucumber";
 import assert      from "node:assert/strict";
 import npmValidate from "validate-npm-package-name";
 
-Given("(that )the {string} NPM package is installed", verifyNPMPackage);
+Given("(that )the {refstr} NPM package is installed", verifyNPMPackage);
```
adamlacoste, line 10: `refstr` is the name I've given for strings that should be parsed for ref notation.  In general, we should use this anywhere we would otherwise want to accept a string, and other string-based param types should piggyback on this logic.  I'm open to changing the name (e.g. `refstring` or `refString`) but since users will need to type this in their own step defs, I thought keeping it short and simple was best.

modules/@specify-bdd/plugin-cli/src/cucumber/support/param_types.ts
adamlacoste, line 1: This file contained only param types that were variations on `ref`, specifically `ref:exitCode` and `ref:terminalOutput`.  Since we've retired those type-specific variations in favor of the new general purpose format, we didn't need this module anymore.

modules/@specify-bdd/quick-ref/src/lib/QuickRef.ts
```diff
@@ -9,8 +9,13 @@ import merge from "deepmerge";
...
 import type { JsonObject, JsonValue } from "type-fest";
 
+export const refNotation = /\$\{[\w.]+\}/;
+
 export class QuickRef {
-    #refs: JsonObject = {};
+    /**
+     * All the references stored in this QuickRef instance.
+     */
+    #all: JsonObject = {};
```
adamlacoste, line 18: I renamed the `refs` private property and its public accessor to `all` because otherwise we were going to have situations where we were doing stuff like `quickRef.refs.ref.whatever` and that seemed real silly.

modules/@specify-bdd/quick-ref/src/lib/QuickRef.ts
```diff
@@ -64,29 +70,41 @@ export class QuickRef {
...
+     *                  will retrieve the value of all.foo.bar.baz.
      *
-     * @returns The value found at given address
+     * @returns The value found at the given address
+     */
+    lookupByAddress(address: string): JsonValue {
+        const keys = address?.split(".") ?? [];
+
+        return this.lookupByKeys(...keys);
+    }
```
adamlacoste, line 90: The `lookup` method expected zero to many params which it would use as lookup keys, but because I was now going to be parsing dot notation, I knew I needed a version of this that handled that parsing.  I kept both versions and have the dot notation form call the key chain form to avoid duplicating logic.

modules/@specify-bdd/quick-ref/src/lib/QuickRef.ts
```diff
@@ -107,4 +125,30 @@ export class QuickRef {
...
+     * addresses found with their corresponding values.
+     * 
+     * @param input - The input string to parse
+     * 
+     * @returns The parsed input string
+     */
+    parse(input: string): string {
+        const matches = input.match(new RegExp(refNotation, "g"));
+
+        if (!matches) {
```
adamlacoste, line 140: `string.match` returns `null` if no matches are found so I didn't want to proceed to the for-of loop below unless we had gotten something back.

modules/@specify-bdd/specify/src/config/paths.config.ts
```diff
@@ -9,4 +11,5 @@ export const paths: PathsConfig = {
     "gherkin": "./features/**/*.feature",
     "logs":    "./logs/specify",
     "refs":    "./specify.refs.json",
+    "rerun":   `${os.tmpdir()}/specify/rerun.txt`,
```
adamlacoste, line 14: Because the temp directory varies from system to system, this ensures that the configured default should work anywhere.

modules/@specify-bdd/specify/src/cucumber/support/hooks.ts
adamlacoste, line 1: Previously, our QuickRef instance was being initialized with the param types as top-level keys; for instance, `exitCode`, `terminalOutput`, etc.  That had to change now that we need to access both ref and config data via the same notation system, so I decided to namespace all ref data under the key `ref` and add `config` alongside that.

modules/@specify-bdd/specify/src/cucumber/support/param_types.ts
```diff
@@ -8,17 +8,18 @@
 import { defineParameterType } from "@cucumber/cucumber";
 import path                    from "node:path";
 
-import type { FileParam } from "~/types/params";
```
adamlacoste, line 11: `FileParam` was not actually being used for anything.

modules/@specify-bdd/specify/src/cucumber/support/param_types.ts
```diff
@@ -8,17 +8,18 @@
...
+import { refNotation } from "@specify-bdd/quick-ref";
+
+import type { JsonValue } from "type-fest";
 
 const ordinal      = /[0-9]*(?:1st|2nd|3rd|[4-90]th)/;
 const quotedString = /"(?:\\.|[^\\"])*"|'(?:\\.|[^\\'])*'/;
-const refName      = /\$\w+/;
```
adamlacoste, line 15: Replaced by `refNotation`

modules/@specify-bdd/specify/src/cucumber/support/param_types.ts
```diff
@@ -8,17 +8,18 @@
...
 
 defineParameterType({
-    "name":   "path",
+    "name":   "filePath",
     "regexp": quotedString,
     transformer(input: string): string {
-        return path.resolve(input.slice(1, -1));
+        return path.resolve(this.quickRef.parse(input.slice(1, -1)));
```
adamlacoste, line 22: It might behoove us to move the standard handling for `refstr` params (that is to say, slice off the quotation marks and then parse as a potential ref string) into some helper module to avoid repeating ourselves.

modules/@specify-bdd/specify/types/params.d.ts
adamlacoste, line 1: This contained the `FileParam` interface which we no longer use.

features/test/rerun.feature
```diff
@@ -7,40 +7,75 @@ Feature: Rerun Failed Tests
...
+    Rule: Rerun does nothing if the last test run passed completely
+
+        @skip @todo
+        Scenario: Rerun after passing test run returns an error
+            Given that the ${ref.file.rerun.path} file content is empty
+            When a user runs the command "npx specify test --rerun --rerun-file=${ref.file.rerun.path}"
+            Then the last command's exit code should be an ${ref.exitCode.error}
+            And the last command's terminal output should match ${ref.terminalOutput.noRerunTestsError}
+
+    Rule: Rerun file path defaults to os.tmpdir()/specify/rerun.txt
```
wild-love, line 58: For a default test like this, I think we need to run some verification about the state of that location.

For example, previous specify execution may have already produced a rerun file in this location.

Additionally, do we want to run destructive operarions on our own rerun file as part of our testing? We may need to copy the file at that location (if it already exists), to another temp location for the test, then move it back.

Then again, we're already running the tests, so maybe its unecessary since it will be overwritten anyway? :thinking:

features/test/rerun.feature
```diff
@@ -7,40 +7,75 @@ Feature: Rerun Failed Tests
...
+
+        @skip @todo
+        Scenario: Failing test run creates a rerun file in the default location
+            When a user runs the command "npx specify test ./binary/failing.feature"
+            Then the ${config.paths.rerun} file path should exist
+            And the ${config.paths.rerun} file content should match ${ref.file.rerun.content.failingScenarios}
```
wild-love, line 64: This will leave a test rerun file at our real rerun location, which we might not want.

wild-love (changes requested):

features/test/rerun.feature
adamlacoste: I would not want to go to the effort of preserving the contents of that rerun file and restoring them after the test, since by its nature it is meant to be disposable.

The initial state of the file when we start the test may vary significantly; the user may have one containing zero-many failed scenarios, or may have no file at all (since it's in a temp location and may be cleared at any time).  Ultimately though, the initial state doesn't matter as much as the state after the tests execute.  That's when we want to verify that it's what we expect, which tells us that we are writing to that location when no `--rerun-file` option is used.

It's technically possible that the initial state of the rerun file could coincidentally match the expected end state, in which case we would not know with certainty that our actions had the intended effect.  If that's a big concern, we can take action to assert an initial state and overwrite or delete whatever's there.  But I think it's a pretty small corner case we're talking about here.

features/test/rerun.feature
```diff
@@ -7,40 +7,75 @@ Feature: Rerun Failed Tests
...
-        Scenario: Rerun is attempted with no prior test run
-            Given that a "failing feature" file exists at "./features"
-            When the user runs the command "npx specify test --rerun"
-            Then the last command's exit code should be an $error
-            And the console output should match $noTestsToRerun
+    Rule: Every test run creates a rerun file
+
+        @skip @todo
+        Scenario: Passing test run creates an empty rerun file
+            When a user runs the command "npx specify test --rerun-file=${ref.file.rerun.path} ./binary/passing.feature"
```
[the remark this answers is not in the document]
adamlacoste: Do you have a recommendation?  We could say "unique temp file path" but that seems cumbersome to me.  I would prefer to explain the uniqueness thing in documentation, personally.

wild-love (approved):

Issue #95, opened 2025-07-16
Automate NPM publishing


adamlacoste:
We need to build GitHub Action automation to publish our packages to NPM when we tag the repo.

PR #170, opened 2025-09-10
Bug fix for workflow variables

adamlacoste:
[#95] Apparently you can't use env vars in the runs-on context, but you can use repo vars; this works out for the better because we can define these values just once in a central location and all workflows now share them

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -2,7 +2,7 @@ name: PR Checks
 on: pull_request
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-${{ vars.UBUNTU_VERSION }}
     steps:
       - name: Check out repository code
         uses: actions/checkout@v4


.github/workflows/checks.yaml
```diff
@@ -1,20 +1,18 @@
 name: PR Checks
 on: pull_request
-env:
-  node_version: 22.x
```
adamlacoste, line 4: Some of the places I was using these `env` vars, such as the `runs-on` context, don't allow access to `env`.  The suggested fix was that I use repo vars instead, found under the `vars` key.  This does mean that we are moving the pinned version of Node and Ubuntu outside the codebase and into the repo configuration, but the upside is that doing so removes the need to define these versions redundantly in every workflow file.  Now when we want to upgrade versions, we just have to change one value in a single location.

.github/workflows/checks.yaml
```diff
@@ -1,20 +1,18 @@
...
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-${{ vars.UBUNTU_VERSION }}
```
adamlacoste, line 5: I wasn't making edits to `checks.yml` originally but upon reviewing it to switch over to using the repo vars, I noticed that this was using `ubuntu-latest` so I fixed that.

.github/workflows/publish.yml
```diff
@@ -11,33 +8,33 @@ permissions:
   packages: write
 jobs:
   # publish-github:
-  #   runs-on: ubuntu-24.04
+  #   runs-on: ubuntu-${{ vars.UBUNTU_VERSION }}
```
adamlacoste, line 11: Inactive but easy to overlook later so I fixed it now.

PR #171, opened 2025-09-10
Implement `--retry-tag`

wild-love:


modules/@specify-bdd/specify/src/config/content.config.ts
```diff
@@ -14,6 +14,8 @@ const contentObj = {
                         "parallel": "Run in parallel with the given number of workers.",
                         "retry":
                             "Retry failed tests up to the given number of times before counting them as failed. (default: %d)",
+                        "retryTag":
+                            "Only retry tests which satisfy this tag expression. (default: %s)",
```
wild-love, line 18: Open to revision suggestions.

modules/@specify-bdd/specify/src/lib/TestCommand.test.ts
```diff
@@ -87,6 +87,21 @@ describe("TestCommand", () => {
                 });
             });
```
wild-love, line 89: I decided to try doing multi-test tests, since theyre closely related. This checks that not only the user arg gets passed to the cucumber config, but also that it overrides any value there that already exists.

modules/@specify-bdd/specify/src/lib/TestCommand.ts
```diff
@@ -37,6 +37,7 @@ export const TEST_COMMAND_DEFAULT_OPTS: TestCommandOptions = {
 
 export interface TestCommandArguments {
     paths?: string[];
+    retryTag?: string;
     tags?: string[];
     watch?: boolean;
 }
```
wild-love, line 43: I havent been updating this when we add new cli options :sweat_smile:

modules/@specify-bdd/specify/src/lib/TestCommand.ts
```diff
@@ -183,7 +187,7 @@ export class TestCommand extends Command {
         }
 
         // Cucumber errors if retry === 0 while retryTagFilter is truthy
-        if (!config.retry) {
+        if (config.retry === 0) {
```
wild-love, line 190: Relatively small change that makes testing slightly less complex (no need to define `retry` arg for the `retryTag` test).

modules/@specify-bdd/specify/src/main.ts
```diff
@@ -59,6 +59,10 @@ app.command("test", { "isDefault": true })
         util.format(helpText.commands.test.options.retry, cucumberCfg.retry),
         parseRetryOption,
     )
+    .option(
+        "-R, --retry-tag <tags>",
+        util.format(helpText.commands.test.options.retryTag, cucumberCfg.retryTagFilter),
+    )
```
wild-love, line 65: No custom parsing for this option, though I can add it if we need to set constraints.

Issue #155, opened 2025-09-04
Commander exit codes


adamlacoste:
Commander currently throws an exception for invalid inputs while parsing arguments, which results in an exit code of 1.  Our intended behavior for specify is to use exit code 1 for test failures and 2+ for error states, meaning something went wrong with the effort to test.  Since invalid command args falls into the second category, we should override the default behavior and ensure an exit code of 2 is returned, even if we have to try/catch the whole command arg parsing effort.

PR #177, opened 2025-09-16
[#155] Implement new exit code for commander errors.

wild-love:


features/test/basic.feature
wild-love, line 1: These are now appropriately errors.

modules/@specify-bdd/specify/src/main.ts
```diff
@@ -49,7 +49,8 @@ app.name("specify")
     .helpCommand("help [command]", helpText.commands.help)
     .helpOption("-h, --help", helpText.options.help);
 
-app.command("test", { "isDefault": true })
+app.exitOverride(() => process.exit(2))
```
wild-love, line 52: This was the key to getting this working; it has to be _before_ the command function, which the official docs actively lead me astray with. See here; 

https://github.com/tj/commander.js/blob/bd4ae263f6966183f3c9e8b105fb5ae3c568c047/tests/command.asterisk.test.js#L79

adamlacoste: Fascinating!

adamlacoste (approved):

PR #178, opened 2025-09-16
NPM publish iteration

adamlacoste:
[#95] Seems I wasn't sending my version output to the right place before

### DIFF
--- a/scripts/github/version.sh
+++ b/scripts/github/version.sh
@@ -6,7 +6,8 @@ source $(dirname $SCRIPT_DIR)/common.sh
 VERSION="$1"
 
 # don't proceed past this point unless we have a VERSION and all the tools we need
-[[ -n "$VERSION" ]] || fail "No version argument supplied."
+[[ -n "$VERSION" ]]       || fail "No version argument supplied."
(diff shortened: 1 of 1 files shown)


scripts/github/version.sh
```diff
@@ -23,6 +24,6 @@ MOD_VTAGS="$(pnpm -r exec -- npm version --no-git-tag-version $VTAG)" \
     || fail "Failed to set the module-level package version."
 debug "Module-level package versions set."
 
-echo "version=$VTAG"
+echo "version=$VTAG" >> "$GITHUB_OUTPUT"
```
adamlacoste, line 27: I was just echoing this to STDOUT but according to the documentation there is some other mysterious location stored in the env var `$GITHUB_OUTPUT` I need to point it toward.

The addition of this change makes this script less testable in a local environment; to do so now, you have to ensure that `GITHUB_OUTPUT` is defined for this script and that it points to a location you don't mind modifying.  (Prepending `GITHUB_OUTPUT=/dev/null` to the script execution command would probably work best.)  I don't anticipate needing to run this script manually under any normal circumstances, but it's worth bearing in mind.

PR #181, opened 2025-09-17
Docs got a point here

adamlacoste:
[#133] Made the address param for QuickRef.lookupByAddress optional

### DIFF
--- a/modules/@specify-bdd/quick-ref/src/lib/QuickRef.ts
+++ b/modules/@specify-bdd/quick-ref/src/lib/QuickRef.ts
@@ -76,10 +76,11 @@ export class QuickRef {
     }
 
     /**
-     * Look up a reference by its dot notation address.
+     * Look up a reference by its dot notation address.  If no address is 
+     * provided, the entire reference store will be returned.
      *
-     * @param address - The address to look up.  For example, ("foo.bar.baz")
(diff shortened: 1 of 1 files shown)


modules/@specify-bdd/quick-ref/src/lib/QuickRef.ts
```diff
@@ -83,7 +83,7 @@ export class QuickRef {
      *
      * @returns The value found at the given address
      */
-    lookupByAddress(address: string): JsonValue {
+    lookupByAddress(address?: string): JsonValue {
```
wild-love, line 86: Do we want to make this optional? Seems like the point of looking up by address is to look up by... an address :thinking:

adamlacoste: I made it optional because the underlying `lookupByKeys` can accept zero arguments, in which case it returns the whole reference store object.  There isn't an especially good reason to do this, but I didn't see any reason not to allow it. 🤷

wild-love: I suppose its ok, just seems a bit odd to me. In that case, I think the behavior should be documented in the docblock somewhere (maybe the return?).

wild-love (approved):

PR #189, opened 2025-09-18
Trying deploy key solution

adamlacoste:
[#95] Added deploy key to checkout action, hopefully this will allow us to bypass branch protections; also moved git configuration out of the commit-tag script and into a separate step, which will make the script less brittle

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,6 +34,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           ref: develop
+          ssh-key: "${{ secrets.RELEASE_DEPLOY_KEY }}"
       - name: Setup PNPM
         uses: pnpm/action-setup@v4
       - name: Setup Node.js ${{ vars.NODE_VERSION }}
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -34,14 +34,15 @@ jobs:
         uses: actions/checkout@v4
         with:
           ref: develop
+          ssh-key: "${{ secrets.RELEASE_DEPLOY_KEY }}"
```
adamlacoste, line 37: This is the substantive change for this PR.  I've configured a deploy key in our repo which should have the ability to bypass branch protections for `develop` and `main`.  If this works the way I intend for it to, the `git push` that happens in the commit-tag script will use this key and be allowed to do what it needs to do.

.github/workflows/release.yml
```diff
@@ -50,16 +51,17 @@ jobs:
         run: pnpm test
       - name: Audit
         run: pnpm audit --prod --audit-level=moderate
+      - name: Configure Git
+        run: |
+          git config user.name "${{ vars.GIT_USER_NAME }}"
+          git config user.email "${{ vars.GIT_USER_EMAIL }}"
```
adamlacoste, line 57: I previously put these lines in the commit_tag.sh script, but doing so introduced the requirement that certain environment vars were defined or else the script couldn't run.  This obviously would make it more of a hassle to run the script in any context other than a github action.  By moving these lines into their own step, I can remove that requirement and let the commit_tag.sh script run without unnecessary strings attached.  If we need to run it manually, it will just act as whichever user is configured for that repo, as one would probably expect.

PR #190, opened 2025-09-18
YAML all the way

adamlacoste:
[#95] OK this is pretty weird but I've tried to capture the whole release process in a single workflow file with no additional scripts called


.github/workflows/release.yml
```diff
@@ -22,18 +22,22 @@ permissions:
...
-      git_tag: ${{ steps.pkg_versions.outputs.version }}
-      title: ${{ steps.commit_tag.outputs.title }}
+      git_tag: ${{ steps.reg.outputs.git_tag }}
+      title: ${{ steps.reg.outputs.title }}
     steps:
       - name: Check out repository code
         uses: actions/checkout@v4
         with:
-          ref: develop
+          ref: ${{ env.GIT_BRANCH }}
```
adamlacoste, line 40: This notation ensures that the branch that we check out and the branch we push back to origin will always be the same one.

.github/workflows/release.yml
```diff
@@ -55,30 +59,55 @@ jobs:
         run: |
           git config user.name "${{ vars.GIT_USER_NAME }}"
           git config user.email "${{ vars.GIT_USER_EMAIL }}"
-      - id: pkg_versions
-        name: Update package versions
-        run: ./scripts/github/version.sh "${{ inputs.version }}"
+      - name: Update monorepo package version
+        run: echo "GIT_TAG=$(npm version --no-git-tag-version ${{ inputs.version }})" >> "$GITHUB_ENV"
```
adamlacoste, line 63: Instead of running the version.sh script, we've cannibalized its contents in two new steps:
1. Update the top-level package.json and capture the semver version that command spits out
2. Update the module-level package.json (but ignore their output, it's redudant).

You'll see in the first step that we have to do some goofy stuff, using the output from the `npm version` command to create an echo line something like `GIT_TAG=v1.2.3` and then directing that output to a location indicated by the variable `GITHUB_ENV`.  This will set `GIT_TAG` for all subsequent steps in this job.

.github/workflows/release.yml
```diff
@@ -55,30 +59,55 @@ jobs:
...
       - name: Generate docs
         run: pnpm docs:all
-      - id: commit_tag
-        name: Commit changes and tag
-        run: ./scripts/github/commit_tag.sh "${{ steps.pkg_versions.outputs.version }}" "${{ inputs.title }}"
+      - name: Set default release title
+        run: echo "TITLE=Specify release $GIT_TAG" >> "$GITHUB_ENV"
+      - name: Append user input title
+        if: ${{ inputs.title }}
+        run: echo "TITLE=$TITLE (${{ inputs.title }})" >> "$GITHUB_ENV"
```
adamlacoste, line 72: This took a little thought and is the part where I was initially most convinced I needed a script to do this.  We'll see if this works!  In a two-step process, we first set a default title for the release and then, conditional on having provided a release title when triggering the workflow, augment that default title with the user-supplied name.

.github/workflows/release.yml
```diff
@@ -55,30 +59,55 @@ jobs:
...
+        run: git commit -a -m "$TITLE"
+      - name: Create release tag ${{ env.GIT_TAG }}
+        run: git tag -a -m "$TITLE" $GIT_TAG
+      - name: Push changes to ${{ env.GIT_REMOTE }}
+        run: git push --tags $GIT_REMOTE $GIT_BRANCH
+      - name: Register job output values
+        id: reg
+        run: |
+          echo "git_tag=$GIT_TAG" >> "$GITHUB_OUTPUT"
+          echo "title=$TITLE" >> "$GITHUB_OUTPUT"
```
adamlacoste, line 83: Every step that is going to produce outputs used by subsequent jobs needs an ID so it can be referenced in the job outputs section nearer to the top.  I figured it would be simpler to just do this as one step at the end, so that all the rest of the step logic can work with env vars and reduce the logical complexity.

.github/workflows/release.yml
```diff
@@ -55,30 +59,55 @@ jobs:
...
     needs: tag
     runs-on: ubuntu-${{ vars.UBUNTU_VERSION }}
+    env:
+      GIT_BRANCH: main
     steps:
       - name: Check out repository code
         uses: actions/checkout@v4
         with:
-          ref: main
+          ref: ${{ env.GIT_BRANCH }}
```
adamlacoste, line 94: Same as before, we check out and push back to the same branch, this time `main`.

.github/workflows/release.yml
```diff
@@ -55,30 +59,55 @@ jobs:
...
           ssh-key: "${{ secrets.RELEASE_DEPLOY_KEY }}"
       - name: Merge develop branch into main
-        run: ./scripts/github/merge_develop_to_main.sh "${{ needs.tag.outputs.git_tag }}"
+        run: git merge --ff-only "${{ needs.tag.outputs.git_tag }}"
```
adamlacoste, line 97: I'm not actually certain whether this will work; I think GitHub is performing a shallow clone on checkout so the tags might not be available.  If that proves to be the case, I'll need to do some additional fetching, but I thought I'd at least try it the simple/lean way first.

.github/workflows/release.yml
```diff
@@ -55,30 +59,55 @@ jobs:
...
       - name: Create release
-        run: ./scripts/github/release.sh "${{ needs.tag.outputs.git_tag }}" "${{ inputs.title }}"
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh release create "${{ needs.tag.outputs.git_tag }}" \
+            --title "${{ inputs.title }}" \
+            --fail-on-no-commits \
+            --generate-notes \
+            --verify-tag
```
adamlacoste, line 113: I hope this notation breaking this command across multiple lines works here, because this one runs real long if I can't do that.

PR #196, opened 2025-09-27
Dev/184/only failures rerun

wild-love:


features/test/rerun.feature
```diff
@@ -30,18 +30,16 @@ Feature: Rerun Failed Tests
 
     Rule: Rerun only executes the tests that failed in the last test run
 
-        @skip @todo
         Scenario: Rerun only one scenario in failing.feature
             Given that the ${rerun.path} file content is "binary/failing.feature:3"
             When a user runs the command "npx specify test --rerun --rerun-file=${rerun.path}"
-            Then the last command's terminal output should match "1 scenario \\(1 failed)"
+            Then the last command's terminal output should match "1 scenario \(1 failed\)"
```
wild-love, line 36: Parentheses need to be escaped for the regex to work, but escape characters dont 🤔

modules/@specify-bdd/specify/src/lib/__mocks__/RerunFile.ts
wild-love, line 1: Super simple mock. I would have liked to use dependency injection in the `TestCommand`, but I couldnt think of a good place for it and we're not injecting `CucumberTool` either, so I figured we'd clean it up in the future.

modules/@specify-bdd/specify/src/lib/tests/RerunFile.test.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    describe("read()", () => {
+        it("reads a file from path", async () => {
+            await RerunFile.read("/foo/bar");
+
+            expect(fs.readFile).toBeCalledWith("/foo/bar", { "encoding": "utf-8" });
+        });
+
+        it("returns an array of paths when reading a rerun file", async () => {
+            vi.mocked(fs).readFile.mockResolvedValueOnce("featureA:1\nfeatureB:2:3\nfeatureC:4");
```
wild-love, line 19: Found this handy thing, `vi.mocked()`. It makes TypeScript happy with mocked modules, so no more red squiggles in the editor!

modules/@specify-bdd/specify/src/lib/RerunFile.ts
wild-love, line 1: Thought I'd break this functionality out into its own class. Seems useful enough on its own, and we might find additional functionality to stuff in here.

specify.refs.json
```diff
@@ -7,7 +7,7 @@
     "file": {
         "rerun": {
             "content": {
-                "failingScenarios": "^binary/failing.feature:3:7:11$"
+                "failingScenarios": "binary/failing.feature:3:7:11$"
```
wild-love, line 10: Rerun paths are absolute now.

features/test/rerun.feature
adamlacoste: 🤷

modules/@specify-bdd/specify/src/lib/RerunFile.ts
```diff
@@ -0,0 +1,41 @@
...
+export class RerunFile {
+    /**
+     * Read the contents of a rerun file.
+     *
+     * @param filepath - The path to the rerun file
+     *
+     * @returns An array of paths contained in the rerun file
+     */
+    static async read(filepath: string): Promise<string[]> {
+        const contents = await readFile(filepath, { "encoding": "utf-8" });
```
adamlacoste, line 16: The [encoding string](https://nodejs.org/docs/latest-v22.x/api/fs.html#filehandlewritefiledata-options) should be `utf8`.  There's another instance of this to be corrected below, as well.

modules/@specify-bdd/specify/src/lib/RerunFile.ts
```diff
@@ -0,0 +1,41 @@
...
+
+        return contents ? contents.split("\n") : [];
+    }
+
+    /**
+     * Convert all relative paths in a rerun file to absolute paths based on the given
+     * absolute base path.
+     *
+     * @param rerunFilepath - The path to the rerun file
+     * @param absolutePath  - The absolute base path to use for conversion
```
adamlacoste, line 26: This param name seems a bit unclear to me.  I suggest any of the following alternatives:
* `absBasePath`
* `basePath`
* `cwd`
* `featureBasePath`
If you disagree and are happy with the current name though, I don't think it's that big a deal and wouldn't withhold approval over this.

modules/@specify-bdd/specify/src/lib/RerunFile.ts
```diff
@@ -0,0 +1,41 @@
...
+     */
+    static async makeAbsolute(rerunFilepath: string, absolutePath: string): Promise<void> {
+        const rerunPaths = await RerunFile.read(rerunFilepath);
+
+        const absolutePaths = rerunPaths.map((rerunPath) => {
+            const separatorIndex = rerunPath.indexOf(":");
+            const featurePath    = rerunPath.slice(0, separatorIndex);
+            const fileLines      = rerunPath.slice(separatorIndex + 1);
```
adamlacoste, line 34: If we didn't add 1 to the index here, we wouldn't need the ":" string to be re-added in the concatenation below.  Not necessarily a problem to do it this way, it's just working harder than it needs to.

modules/@specify-bdd/specify/src/lib/TestCommand.ts
```diff
@@ -132,6 +134,11 @@ export class TestCommand extends Command {
 
             const cucumberRes = await CucumberTool.runCucumber(cucumberRunConfig, cucumberEnv);
 
+            // TODO: remove this conditional once rerun files are always created
+            if(userArgs.rerun && userArgs.rerunFile) {
```
adamlacoste, line 138: No space between `if` and `(` here.  I think the linter enforces this, might have forgotten to run it?

modules/@specify-bdd/specify/src/lib/TestCommand.ts
```diff
@@ -269,4 +284,19 @@ export class TestCommand extends Command {
...
         }
     }
+
+    /**
+     * Get the path that the rerun file will be read from and written to.
+     *
+     * @returns The path found in the formatter, otherwise the default path
+     */
+    #getRerunFilepath(): string {
+        const defaultFilepath = "/tmp/specify/rerun.txt";
```
adamlacoste, line 294: I don't like having this path hard-coded here.  We should be using the `config.paths.rerun` value instead, as that is the permanent home of the default value and is OS-agnostic.  (The above path assumes the existence of a `/tmp` which is not a guarantee.)

Bear in mind though that lib modules shouldn't import config directly; it will need to be one of the values that gets passed in with the `TestCommandOptions` I expect.

adamlacoste (changes requested):
Some changes needed, and some suggestions you're free to take or leave as well.

modules/@specify-bdd/specify/src/lib/RerunFile.ts
wild-love: - [ ] Remove unnecessary concatenation

modules/@specify-bdd/specify/src/lib/RerunFile.ts
wild-love: - [ ] Rename param to `basePath`

modules/@specify-bdd/specify/src/lib/RerunFile.ts
wild-love: - [ ] Change encoding to `utf8`

modules/@specify-bdd/specify/src/lib/TestCommand.ts
wild-love: - [ ] Use the dang linter

modules/@specify-bdd/specify/src/lib/TestCommand.ts
wild-love: This was a temporary measure, Im planning to refactor this into the config file as part of my next ticket, which will deal with the "default value" scenarios.

I can change it now though, if you prefer.

adamlacoste (approved):

PR #201, opened 2025-10-06
Core package readme draft

adamlacoste:


modules/@specify-bdd/specify/package.json
```diff
@@ -49,6 +49,22 @@
...
+    "cucumber",
+    "development",
+    "driven",
+    "e2e",
+    "end to end",
+    "specify",
+    "test",
+    "test-driven development",
+    "testing"
+  ],
```
adamlacoste, line 67: This list is likely incomplete, and it might be that we don't even want to do this until we're sure we're ready for attention on the project.  If that proves to be the case, I'll remove this change from this ticket and restore it in a different branch that can stay on ice until we're ready.

modules/@specify-bdd/specify/README.md
```diff
@@ -1,72 +1,217 @@
...
+| Option                   | Description                                                                        |
+| -------------------------| ---------------------------------------------------------------------------------- |
+| -h, --help               | Display this help screen.                                                          |
+| -p, --parallel <workers> | Run in parallel with the given number of workers.                                  |
+| --rerun                  | Rerun the failed tests from the last test run.                                     |
+| --rerun-file <path>      | The location of the rerun log file.                                                |
+| -r, --retry <retries>    | Retry failed tests up to the given number of times before counting them as failed. |
+| --retry-tag <tags>       | Only retry tests which satisfy this tag expression.                                |
+| -t, --tags <tags>        | Run only the tests which satisfy this tag expression.                              |
+| -w, --watch              | Watch for changes and re-run tests.                                                |
```
adamlacoste, line 163: This list has some changes from the one `specify help test` currently produces:
1. ordering; this is alphabetical, which Commander (frustratingly) doesn't do
2. the `--retry-tag` option is not given the `-R` short opt, which I think we should reserve for other purposes
3. all default notations have been removed; I did this because those defaults might change and I doubt anyone would remember to update this doc when they do, but I can be talked out of this if somebody thinks we need em
4. The language for opt args (e.g. `<number_of_workers>`) has been shortened (`<workers>`) to help everything fit better

We might or might not want to adopt some or all of these changes in our help text as well; please share takes.

modules/@specify-bdd/specify/README.md
```diff
@@ -1,72 +1,217 @@
...
-- Configure runtime behavior via `specify.config.json` in your project's `root` directory
-- Integrate with any WebDriver-compatible browser
+---
 
-### Custom Step Definitions
+## Get Help
 
-**I AM A STUB, FIX ME @adamlacoste**
+Having trouble with Specify?  We're here to help.  Click
+[here](https://github.com/specify-bdd/specify-core/discussions/new?category=q-a) to open a new Q&A discussion topic on
```
adamlacoste, line 216: This works!

PR #202, opened 2025-10-07
CLI readme

adamlacoste:


modules/@specify-bdd/plugin-cli/README.md
```diff
@@ -0,0 +1,62 @@
+# Specify BDD - CLI Plugin
+
+This plugin for [Specify BDD](https://www.npmjs.com/package/@specify-bdd/specify) provides a bundle of Cucumber support
+code designed to make testing CLI commands in a shell simple.  Please refer to the core package's README file for 
+instructions on installation and configuration.
+
+## Cucumber Support Code
```
adamlacoste, line 7: Probably goes without saying but I want to replace everything listed below with some kind of auto-generated lists in the next release.

modules/@specify-bdd/specify/README.md
adamlacoste, line 1: These horizontal rules in the core package readme proved to be both unnecessary and ugly given the way GitHub renders the markdown, so I pulled out all but the first one separating the TOC from the main content.

Issue #210, opened 2025-10-09
Timing failure in parallel feature testing


adamlacoste:
One of the scenarios of the parallel execution feature fails intermittently when I run it on my laptop.  It's an older model from 2016 and likely runs slower than the average bear.  We do not appear to have the same problem in our GitHub action runners, which seem to pass consistently, but ideally we want our test suite to be able to pass on any machine that can run it, even if it runs slowly.

The root of the problem is that we're using timing to determine whether we're running tests in parallel, which is a brittle tool for that job.  We need to come up with an alternative means of testing parallel processing that actually counts workers rather than depending on execution time.

```
Failures:

1) Scenario: Parallel execution takes less than 4 seconds # features/test/parallel.feature:21
   ✔ Before (CLI plugin before hook) # file:/Users/adam/Development/specify-bdd/core/modules/@specify-bdd/plugin-cli/dist/cucumber/support/hooks.js:12
   ✔ Before (Core before hook) # file:/Users/adam/Development/specify-bdd/core/modules/@specify-bdd/specify/dist/cucumber/support/hooks.js:22
   ✔ Given that the "@specify-bdd/specify" NPM package is installed # file:/Users/adam/Development/specify-bdd/core/modules/@specify-bdd/plugin-cli/dist/cucumber/step_definitions/packages.js:9
   ✔ And that the "@specify-bdd/plugin-cli" NPM package is installed # file:/Users/adam/Development/specify-bdd/core/modules/@specify-bdd/plugin-cli/dist/cucumber/step_definitions/packages.js:9
   ✔ And a CLI shell # file:/Users/adam/Development/specify-bdd/core/modules/@specify-bdd/plugin-cli/dist/cucumber/step_definitions/shell.js:11
   ✔ And that the working directory is "./assets/gherkin" # file:/Users/adam/Development/specify-bdd/core/modules/@specify-bdd/plugin-cli/dist/cucumber/step_definitions/file_system.js:13
   ✔ When a user runs the command "npx specify test --parallel 2 ./slow.feature" # file:/Users/adam/Development/specify-bdd/core/modules/@specify-bdd/plugin-cli/dist/cucumber/step_definitions/shell.js:14
   ✔ Then the last command's exit code should be a ${ref.exitCode.success} # file:/Users/adam/Development/specify-bdd/core/modules/@specify-bdd/plugin-cli/dist/cucumber/step_definitions/shell.js:31
[...]

PR #213, opened 2025-10-14
Just a little parallel test tweak nbd

adamlacoste:
[#210] this started out as a simple change to try to clarify how parallel testing works, as it originally depended on the timing of two slow scenarios, which was predictably brittle; in enabling a new step def capable of detecting how many workers cucumber has running, it was necessary to tweak some things about how ShellSession spawns child processes to ensure the parent's cucumber settings didn't confuse a child shell process running Specify; also started a reorg effort to move assets into a test sandbox directory, but the work is unfinished because the testing step defs can't move into the sandbox or they don't get built TS-transpiled; we'll have to circle back to that in a later release and finish the job

### DIFF
--- a/modules/@specify-bdd/plugin-cli/src/lib/SessionManager.ts
+++ b/modules/@specify-bdd/plugin-cli/src/lib/SessionManager.ts
@@ -44,7 +44,7 @@ interface Delimiter {
  */
 interface DelimiterPayload {
     cwd: string;
-    exitCode: string;
+    exitCode: string; // this is initially set with a util.format placeholder (%d) so it must be a string type
     uuid: string;
 }
 
@@ -393,8 +393,8 @@ export class SessionManager {
         const suffix = "SPECIFY DELIM END";
         const uuid   = randomUUID(); // used to prevent false delimits
 
-        const payload: DelimiterPayload = { "exitCode": "%d", "cwd": "%s", uuid };
-        const payloadStr                = JSON.stringify(payload);
+        const payload    = { "exitCode": "%d", "cwd": "%s", uuid } as DelimiterPayload;
+        const payloadStr = JSON.stringify(payload);
 
         return {
             "command": `;printf '${prefix} ${payloadStr} ${suffix}' "$?" "$(pwd)"`, // Windows friendly?


features/plugin-cli/steps/file_system.feature
```diff
@@ -11,5 +11,5 @@ Feature: File System Step Definitions
             And a CLI shell
 
         Scenario:
-            When the user changes the working directory to "./assets"
+            When the user changes the working directory to "./test"
```
adamlacoste: All assets have moved into a new directory, `./test`, which is meant to serve as an all-purpose Specify testing sandbox going forward.  It currently fills the same function `./assets` did, but its use as a working directory for tests (so as to avoid config pollution at the monorepo top level) is hopefully more obvious this way.  We're still considering how to make this whole arrangement work and might change this further.

features/test/parallel.feature
```diff
@@ -7,21 +7,19 @@ Feature: Parallel Execution
...
         And a CLI shell
-        And that the working directory is "./assets/gherkin"
+        And that the working directory is "./test"
 
     Rule: Tests run serially by default
 
-        Scenario: Serial execution takes more than 4 seconds
-            When a user runs the command "npx specify test ./slow.feature"
+        Scenario: Serial execution starts only one worker
+            When a user runs the command "npx specify test ./gherkin/parallel/workers1.feature"
```
adamlacoste: These were the tests that were flaky before, subject to race conditions based on the host system's speed.  (They failed frequently on my old ass laptop.)  Because timing was an imperfect measure of parallelism, I created new gherkin scenarios and supporting a step def that validates the number of parallel workers more directly.

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
```diff
@@ -131,7 +133,22 @@ async function execCommandSync(command: string): Promise<void> {
...
-    const shell = new ShellSession(this.parameters.userPath);
+    const options: SpawnOptions = { "env": process.env };
+
+    // strip Cucumber env vars from the options object that will be passed to the child process
+    // helps to ensure a Specify process run by Specify doesn't get confused by its parent's operating parameters
+    for (const key in options.env) {
+        if (key.slice(0, 9) === "CUCUMBER_") {
+            delete options.env[key];
+        }
+    }
```
adamlacoste: This is new logic which manipulates the copy of the parent's `process.env` object before we pass it to the child process, so as to prevent the child from inheriting the details of the parent's parallelism settings.

modules/@specify-bdd/specify/src/cucumber/step_definitions/testing.ts
```diff
@@ -6,40 +6,40 @@
...
  */
-async function passAfterDelay(delay: number): Promise<void> {
-    return new Promise((resolve) => setTimeout(resolve, delay * 1000));
+async function passIfNWorkers(workers: number): Promise<void> {
+    assert.equal(
+        parseInt(process.env.CUCUMBER_TOTAL_WORKERS, 10) || 1,
+        workers,
+        "Number of parallel workers did not match expectations.",
+    );
 }
```
adamlacoste: Here's the new definition, which uses the `CUCUMBER_TOTAL_WORKERS` env var set by Cucumber when running in parallel mode to determine if Specify is running with the correct parallelism.  Note however that this env var is NOT set my Cucumber if running in serial mode, which is the reason for that `|| 1` at the end of the actual expression; if the env var isn't set, we take that to mean that we're running in serial.

As mentioned above, this got tricky when running Specify in Specify because the inner process was inheriting the outer's env vars, and the outer Specify is configured to run in parallel.  That's NBD if inner Cucumber is also running in parallel because it overwrites the value of `CUCUMBER_TOTAL_WORKERS` in its internal environment, but if it's running in serial mode, it doesn't set that env var at all and the inherited parent value was allowed to stand, messing up efforts to test it.  I solved this by stripping env vars starting with `CUCUMBER_` from the env inheritance and that appears to have solved the issue.

features/test/watch.feature
```diff
@@ -7,75 +7,73 @@ Feature: Watch Mode
...
-        And that the working directory is "./assets/gherkin"
+        And that the working directory is "./test"
 
     Rule: Watch mode monitors file changes and reruns tests automatically
 
         Scenario: Watch mode starts and monitors for changes
-            When a user starts the async command "npx specify test --watch ./binary/passing.feature"
-            And a user waits for terminal output matching "Watching for changes"
+            When a user starts the async command "npx specify test --watch ./gherkin/binary/passing.feature"
+            And the user waits for terminal output matching "Watching for changes"
```
adamlacoste: This `a/the` change is common throughout this document.  Here's my rationale: if we keep saying "a user" on line after line, it reads to me like each action could be taken by any user; it's not clear grammatically that it's all the same person.  Making each line after the first one "the user" instead sorta fixes this.

Could it be better?  Absolutely.  I really want to make these sentence subjects implicit in a future update so we could just chain off of the subject of the previous step; I think the language would flow much better that way.  We could also just go with "the user" universally throughout our specs so that we aren't switching articles, but because that involved a bigger change to more files, I didn't go that route.

Incidentally, I wondered as I was reading these if we had this problem anywhere else, and to my surprise the rest of the feature set is very consistent in using just one `when` step per scenario.  That's admirable!  The fact that we can't get away with that here makes me wonder if we should reconsider how we're testing this, but that's an optimization for later.

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/file_system.ts
```diff
@@ -9,23 +9,20 @@ import assert, { AssertionError               } from "node:assert/strict";
...
-Given("(that )the {ref} file content is {string}", setFileContent);
+Given("(that )the {filePath} file content is {string}", writeFileContent);
+Given("(that )the {ref} file content is {string}", writeFileContent);
 
-Given("(that )the {filePath} file content is empty", createEmptyFile);
-Given("(that )the {ref} file content is empty", createEmptyFile);
+Given("(that )the {filePath} file content is empty", writeEmptyFileContent);
+Given("(that )the {ref} file content is empty", writeEmptyFileContent);
 
-When("a/the user changes the working directory to {filePath}", changeDirectory);
```
adamlacoste: The `changeDirectory` handler and its associated step definitions have moved to the Shell module, because they are really about changing the working directory of the active shell.  (They now _also_ change the effective working directory of the scenario for file system's purposes, but that's a convenience measure meant to reduce the confusion of working with shells and file system operations in the same scenario, not the primary purpose of those step defs.)

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
```diff
@@ -125,6 +155,8 @@ function execCommand(command: string): void {
 async function execCommandSync(command: string): Promise<void> {
     execCommand.call(this, command);
     await waitForCommandReturn.call(this);
+
+    this.cli.cwd = this.cli.manager.cwd;
```
adamlacoste: This will be repeated in a lot of the functions here, as we have to keep the scenario's CWD record up to date.

modules/@specify-bdd/plugin-cli/src/lib/SessionManager.ts
```diff
@@ -32,7 +32,19 @@ export interface CommandMeta {
...
+    uuid: string;
+}
+
+/**
+ * Internal structure used to form the payload data encoded as JSON in a command
+ *  delimiter string.
+ */
+interface DelimiterPayload {
+    cwd: string;
+    exitCode: string;
```
adamlacoste: It's important that this property needs to be a string because when we prep the delimiter object, we will be using `printf` format specifiers such as `%d` in place of the real values.  This allows us to sub in the real values on the command line and avoid some of the brittleness issues of representing JSON (with all its quote characters) as a string in a CLI.  The exit code string gets converted back to an integer as part of `#processOutput`.

modules/@specify-bdd/plugin-cli/src/lib/SessionManager.ts
```diff
@@ -365,31 +389,46 @@ export class SessionManager {
      * @returns the new delimiter
      */
     #createDelimiter(): Delimiter {
+        const prefix = "SPECIFY DELIM START";
+        const suffix = "SPECIFY DELIM END";
         const uuid   = randomUUID(); // used to prevent false delimits
-        const prefix = "SPECIFY";
+
+        const payload: DelimiterPayload = { "exitCode": "%d", "cwd": "%s", uuid };
```
adamlacoste: Here are those string format specifiers I mentioned before.  They get replaced with the real value when this command runs in the shell.  If `exitCode` were a number, it obviously couldn't have `%d` as a value and still be valid JSON.

modules/@specify-bdd/plugin-cli/src/lib/SessionManager.ts
```diff
@@ -365,31 +389,46 @@ export class SessionManager {
...
+        const payloadStr                = JSON.stringify(payload);
 
         return {
-            "command": `;echo "${prefix} ${SessionManager.#exitCodeKey}=$? UUID=${uuid}"`,
-            "regexp":  new RegExp(`${prefix} ${SessionManager.#exitCodeKey}=(\\d+) UUID=${uuid}`),
+            "command": `;printf '${prefix} ${payloadStr} ${suffix}' "$?" "$(pwd)"`, // Windows friendly?
```
adamlacoste: Instead of `echo`, we're now using `printf` to print the delimiter and all its JSON goodness.  Why?  Because trying to `echo` a JSON string which contains numerous quotation marks would be hell.  Either you have to escape them all or you have to use single quotes, and if you do the latter, this shell expansion behavior for things like `$?` or `$(pwd)` wouldn't work.

`printf`, on the other hand, lets us provide the JSON string and surrounding delimiters as a template, single quoted, and put the shell-interpreted special values in double quotes as additional arguments.  It substitutes those values into the template in the sequence that the format specifiers appear in the string.

I am proud of the solution as a fix for the knotty issues of managing JSON in the CLI, but I'm a little concerned about portability.  `printf` is based on an old C utility and is likely to have nieces and nephews on pretty much any modern system, but will they work enough alike for this command to work?  I worry that a lot of our delimiter stuff is very Linux-centric and some of these assumptions might break down once we take it on the road.  I suppose we'll just have to solve those issues as we discover them though.

modules/@specify-bdd/plugin-cli/src/lib/SessionManager.ts
```diff
@@ -365,31 +389,46 @@ export class SessionManager {
...
+        assert.ok(match, "Output does not contain a matching delimiter.");
 
-        assert.ok(match, `Output does not contain a value for "${key}"`);
+        const payload: DelimiterPayload = JSON.parse(match[1]);
 
-        return parseAsInt ? parseInt(match[1], 10) : match[1];
+        assert.equal(payload.uuid, delimiter.uuid, "Output delimiter is malformed.");
+
+        return payload;
     }
```
adamlacoste: Fairly straightforward replacement, the main differences are 1) the new logic uses the `regexp` we already store on the delimiter anyway to do the matching; and 2) the integer parsing is handed off to `#processOutput`.

I have an assertion in here requiring that the UUIDs of the delimiter and payload match; this is my way of guaranteeing that the parsed JSON is the right stuff.  But that said, it might be a little redundant because as you'll see in `#processOutput`, this method doesn't even get called unless the UUID is detected within the output.  The only edge case this assertion is really protecting us against is one where a chunk of output includes the delimiter's UUID, but also happens to include multiple complete delimiter strings.  Only the first one would be matched, parsed, and compared.  If we don't get the UUID back we expect, kaboom.  I'm not sure we need to be doing this, but I also couldn't think of a good reason not to check since we have the values ready at hand anyway.

.gitignore
```diff
@@ -1,6 +1,7 @@
 # Logs
 logs
 *.log
+test/watch-test-file.txt
```
adamlacoste: I don't think this file belongs in the repo.  It's just testing trash.

.gitignore
```diff
@@ -1,6 +1,7 @@
 # Logs
 logs
 *.log
+test/*.txt
```
adamlacoste: So we don't accidentally add to the repo any testing trash we generate in the course of testing.

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/file_system.ts
```diff
@@ -85,20 +51,11 @@ async function createTempFileRef(address: string): Promise<void> {
  * @param filePath - The path to the file
  */
 async function deleteFile(filePath: string): Promise<void> {
-    await verifyFilePathExists.call(this, filePath);
+    const resolvedPath = resolve(this.cli.cwd, filePath);
```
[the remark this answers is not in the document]
wild-love: Because this is repeated so often, I think we ought to extract it as a helper, either standalone in this step def module, or as part of the world object.

- [ ] Extract path resolver as helper function

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
```diff
@@ -103,6 +109,31 @@ Then(
...
+ *
+ * @throws AssertionError
+ * If the `cd` command returns a non-zero exit code.
+ */
+async function changeDirectory(dirPath: string): Promise<void> {
+    assert.ok(this.cli.manager, new AssertionError({ "message": "No shell session initialized." }));
+
+    const resolvedPath = path.resolve(this.cli.cwd, dirPath);
+
+    this.cli.manager.run(`cd ${resolvedPath}`);
```
wild-love: Couldnt we just await this, instead of awaiting on line 130?

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
wild-love: Hmm... could we use this this method of avoiding `env` cross contamination to also avoid config pollution, and remove the necessity to cd into another directory to run Specify in Specify?

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/file_system.ts
```diff
@@ -9,23 +9,20 @@ import assert, { AssertionError               } from "node:assert/strict";
...
-import { join                                 } from "node:path";
+import { join, resolve                        } from "node:path";
 
 Given("a new temp file path referenced as {string}", createTempFileRef);
-Given("(that )the working directory is {filePath}", changeDirectory);
 
-Given("(that )the {filePath} file content is {string}", setFileContent);
-Given("(that )the {ref} file content is {string}", setFileContent);
+Given("(that )the {filePath} file content is {string}", writeFileContent);
+Given("(that )the {ref} file content is {string}", writeFileContent);
```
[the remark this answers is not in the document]
adamlacoste: Because the file may or may not already exist.  "Write" works either way.  (I imagine that's why Node went with it.)

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/file_system.ts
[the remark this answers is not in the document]
adamlacoste: Something else occurs to me, if you want to entertain the notion: we could have this path resolution happen implicitly as part of the `filePath` param type definition's transformer method.  Those functions have access to the World object via `this` so that could do the work in a single location.  My only reservation about it is how it kinda hides that logic away in a way that might be harder to understand later after we forget how all this works. 🤷

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
adamlacoste: `SessionManager.run()` is not an async method; it only kicks off the command in the shell and returns the `CommandMeta`, it doesn't hang around to see how it turns out.  We made asynchronous operation the default behavior so every place we `run`, we also have to subsequently `waitForReturn`.

modules/@specify-bdd/plugin-cli/src/lib/SessionManager.ts
adamlacoste: - [ ] Add comment explaining `DelimiterPayload`'s string property `exitCode`

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
wild-love: Ah, I see. I must be remembering a previous implementation 😅

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/file_system.ts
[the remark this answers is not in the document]
adamlacoste: - [ ] Move scenario CWD path resolution into filePath param type def

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
adamlacoste: That config manipulation would have to happen within the child process (specify in specify), since that's where the config loading happens. It isn't like the env manipulation where the parent modifies the child's env. And because specify doesn't (and shouldn't?) know whether it is a child process, it couldn't happen ONLY in child processes. This sounds pretty hacky to me, the only ways I can think to make it work seem much more drastic and complicated than a dir change.

wild-love (approved):

Issue #135, opened 2025-08-18
Feature: Watch mode


adamlacoste:
Implement watch mode, which executes tests when changes to the application code are detected, as documented in runner/features/specify-watch.feature.  (We should rename this file to just "watch.feature".)

PR #216, opened 2025-10-16
[#135][#208] Implement kill signal watch scenario.

wild-love:


features/test/watch.feature
```diff
@@ -62,20 +62,20 @@ Feature: Watch Mode
...
 
-        @skip
         Scenario: Watch mode exits with error status when interrupted during test run
-            Given that a "long running feature" file exists at "./features"
-            And that watch mode is running
-            When the user sends a SIGINT signal during test execution
-            Then the watch mode should stop gracefully
-            And the command should exit with an appropriate status code
+            When a user starts the async command "npx specify test --watch ./slow.feature"
+            And a user waits for 0.5 seconds
```
wild-love: Wait here so we know we're in the thick of the 2 second test.

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
```diff
@@ -125,6 +127,16 @@ async function execCommandSync(command: string): Promise<void> {
     await waitForCommandReturn.call(this);
 }
 
+/**
+ * Send a system kill signal to the command in the last used CLI.
+ *
+ * @param signal - The system signal to pass to killCommand()
+ */
+async function sendKillSignal(signal: string): Promise<void> {
```
wild-love: This might need to be renamed later if we want to have an additional step def for killing the session.

PR #220, opened 2025-11-03
Node.js LTS testing

### LINKED ISSUE #219: Automate testing in multiple Node LTS versions
Current automation builds & tests the code under a Node LTS version defined as a repo variable.  We need to expand this system to encompass testing in all currently maintained Node LTS versions to ensure compatibility, and it would probably behoove us at this point to overhaul the automation to reuse artifacts across jobs so we can reduce waste.

adamlacoste:
We need to start testing multiple Node.js versions to ensure compatibility with all active/maintenance LTS releases.  Our release automation can parallelize this process so that it shouldn't take too much longer than the current approach, and we can take advantage of this opportunity to overhaul some inefficiencies in the release workflow (e.g. reducing the amount of redundant building of dist files with artifact preservation).


.github/workflows/release-stage-1-test.yml
adamlacoste, line 1: The addition of a new release workflow stage (test) at the front of the process really created a bunch of weirdness throughout these scripts.  Their names include numbers to make it clear what sequence they execute in, which is nice enough but means a lot of disruption when you have to mess with the order of operations.  The new stage cannibalizes some of the logic from the tag stage (now number 2 in order) to install and build dependencies, and in order to minimize redundancy and potential for divergence, I'm now preserving the built code generated by the test stage for reuse by later stages.

In this new test stage, I'm running the same reusable workflow three times in two separate jobs.  The first one (`test-active`) runs it for the active LTS version of Node and passes in a flag to store the generated artifacts for use by later stages; this ensures that the version of the built code we distribute is always the one generated with the latest LTS.  The second job (`test-maintenance`) uses that workflow for two older LTS versions but doesn't store the built code.

.github/workflows/release-stage-1a-test-node-version.yml
adamlacoste, line 1: Yes, stage 1A.  I didn't like breaking it down this way but the alternative was to repeat this same code twice in the above stage 1 script with only the minor variation of one additional step at the end.  That seemed sloppy and error prone, so I abstracted it out, hence the awkward name.

This workflow PNPM installs dependencies and builds, then runs unit and e2e tests.  Because of the logic in the stage 1 script above, this workflow will run multiple times in parallel for every version on Node.js we want to support.  For this reason, I didn't include the `pnpm lint` or `pnpm audit` actions here since they don't need to be run redundantly for different Node.js versions.

.github/workflows/release-stage-2-tag.yml
adamlacoste, line 1: This stage has had its `pnpm install` and `pnpm test` steps stripped out and replaced by a new one that downloads the build artifacts generated by stage 1.

.github/workflows/release-stage-5-publish.yml
adamlacoste, line 1: Apart from changes here similar to the tag stage script, adapting this stage to use the generated build artifacts instead of re-generating them redundantly, I activated the GitHub Packages publish job which I had completely forgotten I had in here.

modules/@specify-bdd/specify/package.json
```diff
@@ -66,6 +66,6 @@
     "testing"
   ],
   "engines": {
-    "node": ">=22.0.0"
+    "node": ">=20.19.5"
```
adamlacoste, line 69: Updated to reflect the automated testing.  We still need to manually update this when we change the versions we support though.

wild-love (approved):

PR #221, opened 2025-11-04
Leaner release workflow

adamlacoste:
We're currently building our typescript four separate times during the release workflow, due to the use of the `prepare` hook to attach that action to every install and publish.  By decoupling these actions, we can make our automation leaner, faster, and less wasteful.


.github/workflows/checks.yaml
adamlacoste, line 1: With these changes we're making to the NPM prepare actions, the PR checks workflow needed some adjustment.  I could have used `dev:install` here but I thought it would be better for the automation to not use dev shortcuts, since we might change the behavior of those to further optimize dev convenience later without remembering where else they're used.

While I was in here, I went ahead and restructured this workflow to make the lint, unit test, and e2e test jobs separate and parallel rather than serializing everything.  It ought to be faster now as a result.

.github/workflows/release-stage-1a-test-node-version.yml
adamlacoste, line 1: The `workflow_dispatch` section I removed from this script allows the workflow to be triggered manually from the GitHub UI.  Because I can already trigger its parent Stage 1 script, I thought it was unnecessary to also be able to trigger this one directly.

.github/workflows/release-stage-2-tag.yml
adamlacoste, line 1: The inputs required when manually triggering this Stage 2 script gave you a drop down list of options, redundant with the list offered when triggering the whole release process.  Since direct manual activation of this stage script would only happen under unusual circumstances and I could give us greater flexibility and less redundant input definition this way, I replaced the multiple choice input with a string input.

.github/workflows/release.yml
adamlacoste, line 1: We can now create pre-release tags with this automation.  If we were on version `1.2.3` and I chose to "preminor" here, the new tag generated would be `v1.3.0-rc.0`.  When we subsequently generate a "minor" tag, it would be `v1.3.0`.

Because we don't want to be publishing our release candidates publicly (I think?), I have added a conditional to stages 3-5 which prevent them from firing if a pre-release version value was chosen when the workflow was triggered.

package.json
```diff
@@ -12,13 +12,14 @@
...
     "build": "pnpm -r --if-present run build",
+    "dev:i": "pnpm dev:install",
+    "dev:install": "pnpm install && pnpm generate-types && pnpm build",
     "dev:reset": "bash ./scripts/dev-reset.sh",
     "dev:watch": "tsx scripts/dev-watcher.ts",
     "generate-docs": "pnpm -r --if-present run generate-docs",
     "generate-types": "pnpm -r --if-present run generate-types",
     "lint": "prettier --check '**/*.ts' & eslint .",
     "lint:fix": "prettier --write '**/*.ts' && eslint . --fix",
-    "prepare": "pnpm generate-types && pnpm -r --if-present run build",
```
adamlacoste, line 21: Be Prepared

.github/workflows/checks.yaml
adamlacoste: So I just checked and it isn't actually appreciably faster; because the lint and unit tests are SO fast, they were adding only a trivial amount of extra time to the critical path tasks for the e2e testing.  By parallelizing them, I've only shaved a couple seconds off the e2e testing time while adding a bunch of setup/teardown overhead to the lint and unit test jobs.  This approach is actually less efficient, dollars to donuts.

There are still a couple reasons we might want to keep it parallelized.  The parallel PR checks now get unique lines describing their progress on the PR, so if something fails it will be easier to tell what went wrong.  We also don't need to worry much about being inefficient with our github action minutes (for now) because we're not even coming close to spending enough to get charged.

I'm open to keeping or reversing the changes here relating to parallelizing the PR checks.

wild-love (approved):

PR #234, opened 2025-12-04
CucumberManager class

### LINKED ISSUE #224: Create CucumberManager class
We need a new `CucumberManager` class which will wrap around Cucumber and help ensure we're using it in a consistent and orderly way.  The constructor should expect a Cucumber instance as an input param, which will subsequently be stored in a public property on the manger called `Cucumber`.

adamlacoste:
This is just the bare bones of a new `CucumberManager` class and an accompanying unit test for the same.  Follow up tickets will add features to this.

* [#224] Created new CucumberManager class and accompanying unit test

### DIFF
--- a/modules/@specify-bdd/specify/src/lib/CucumberManager.ts
+++ b/modules/@specify-bdd/specify/src/lib/CucumberManager.ts
@@ -1,13 +1,9 @@
 import Cucumber from "@cucumber/cucumber";
 
 export interface CucumberLike {
-    Given: CucumberStepDefLike;
-    Then: CucumberStepDefLike;
-    When: CucumberStepDefLike;
-}
-
-interface CucumberStepDefLike {
-    (pattern: string, options: object, fn: () => void): void;
+    Given: typeof Cucumber.Given;
+    When: typeof Cucumber.When;
+    Then: typeof Cucumber.Then;
 }
 
 let instance: CucumberManager;
--- a/modules/@specify-bdd/specify/src/lib/CucumberManager.ts
+++ b/modules/@specify-bdd/specify/src/lib/CucumberManager.ts
@@ -1,0 +1,46 @@
+import Cucumber from "@cucumber/cucumber";
+
+export interface CucumberLike {
+    Given: CucumberStepDefLike;
+    Then: CucumberStepDefLike;
+    When: CucumberStepDefLike;
+}
+
+interface CucumberStepDefLike {
+    (pattern: string, options?: object, fn: () => void): void;
+}
+
+let instance: CucumberManager | null = null;
+
+/**
+ * A wrapper for Cucumber which streamlines, enhances, and enforces consistency
+ * of Cucumber usage throughout Specify.
+ */
+export class CucumberManager {
+    /**
+     * The managed Cucumber object.
+     */
+    cucumber: CucumberLike;
+
+    /**
+     * Accept a Cucumber object and store it for easy access.
+     *
+     * @param cucumber - The Cucumber object to manage
+     */
+    constructor(cucumber: CucumberLike) {
+        this.cucumber = cucumber;
+    }
+
+    /**
+     * Get the singleton instance of CucumberManager.
+     *
+     * @returns The CucumberManager instance
+     */
+    static getInstance(): CucumberManager {
+        if (instance) {
+            return instance;
+        }
+
+        return (instance = new CucumberManager(Cucumber));
+    }
+}
--- a/modules/@specify-bdd/specify/src/lib/CucumberManager.ts
+++ b/modules/@specify-bdd/specify/src/lib/CucumberManager.ts
@@ -10,7 +10,7 @@ interface CucumberStepDefLike {
     (pattern: string, options?: object, fn: () => void): void;
 }
 
-let instance: CucumberManager | null = null;
+let instance: CucumberManager;
 
 /**
  * A wrapper for Cucumber which streamlines, enhances, and enforces consistency
@@ -37,10 +37,10 @@ export class CucumberManager {
      * @returns The CucumberManager instance
      */
     static getInstance(): CucumberManager {
-        if (instance) {
-            return instance;
+        if (!instance) {
+            instance = new CucumberManager(Cucumber);
         }
 
-        return (instance = new CucumberManager(Cucumber));
+        return instance;
     }
 }
--- a/modules/@specify-bdd/specify/src/lib/CucumberManager.ts
+++ b/modules/@specify-bdd/specify/src/lib/CucumberManager.ts
@@ -7,7 +7,7 @@ export interface CucumberLike {
 }
 
 interface CucumberStepDefLike {
-    (pattern: string, options?: object, fn: () => void): void;
+    (pattern: string, options: object, fn: () => void): void;
 }
 
 let instance: CucumberManager;
--- a/modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
+++ b/modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
@@ -1,0 +1,41 @@
+import { CucumberManager } from "../CucumberManager";
+
+import type { CucumberLike } from "../CucumberManager";
+
+const cucumber: CucumberLike = {};
+// const notCucumber            = null;
+
+describe("CucumberManager", () => {
+    // describe("constructor", () => {
+    //     it("accepts a Cucumber-like object", () => {
+    //         let cm = new CucumberManager(cucumber);
+    //     });
+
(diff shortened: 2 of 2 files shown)


modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
```diff
@@ -0,0 +1,41 @@
...
+describe("CucumberManager", () => {
+    // describe("constructor", () => {
+    //     it("accepts a Cucumber-like object", () => {
+    //         let cm = new CucumberManager(cucumber);
+    //     });
+
+    //     it("rejects an object unlike Cucumber", () => {
+    //         expect(() => new CucumberManager(notCucumber)).toThrow("Invalid input.");
+    //     });
+    // });
```
adamlacoste: These tests have been disabled pending a discussion on the proper way to handle type checking in transpiled JS.  I believe we need some kind of input validation since it is my intention for the Specify package to export this class, but we need to come to some consensus on the best way to do it while minimizing the redundancy of our TS and JS type validation logic.

modules/@specify-bdd/specify/src/lib/CucumberManager.ts
```diff
@@ -0,0 +1,46 @@
+import Cucumber from "@cucumber/cucumber";
+
+export interface CucumberLike {
```
adamlacoste, line 3: I'm not set on this name but it seemed better than any of the alternatives I could come up with. 🤷

modules/@specify-bdd/specify/src/lib/CucumberManager.ts
```diff
@@ -0,0 +1,46 @@
...
+    Given: CucumberStepDefLike;
+    Then: CucumberStepDefLike;
+    When: CucumberStepDefLike;
+}
+
+interface CucumberStepDefLike {
+    (pattern: string, options?: object, fn: () => void): void;
```
adamlacoste: The actual function signature of the step def handler allows for varying input params and return type, so maybe this is not wholly accurate; I'll refine it in future tickets if it gives me trouble.

modules/@specify-bdd/specify/src/lib/CucumberManager.ts
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    /**
+     * Get the singleton instance of CucumberManager.
+     *
+     * @returns The CucumberManager instance
+     */
+    static getInstance(): CucumberManager {
+        if (!instance) {
+            instance = new CucumberManager(Cucumber);
```
adamlacoste, line 37: This `getInstance` static method creates an instance with the `Cucumber` object as input if one doesn't already exists.  This violates a strict dependency injection design pattern, but I think it's worth the tradeoff for the following reasons:
1. It's easier to use; if `getInstance` required a `Cucumber` argument every time it was used, it would require every consumer module to import `Cucumber` as well, and mostly to no avail because we only actually need `Cucumber` for the first call.
2. It's safer to use, because it reduces the chance of different instances of Cucumber being used by different modules; by encapsulating the Cucumber interactions here in this one class module, we ensure everybody who does this the right way gets the same Cucumber instance.
3. The impact to testing is minimal because _only_ the `getInstance` method uses the essential dependency; the constructor and any methods that depend on it can be tested with dependency injection best practice in place.

modules/@specify-bdd/specify/src/lib/CucumberManager.ts
```diff
@@ -0,0 +1,46 @@
+import Cucumber from "@cucumber/cucumber";
+
+export interface CucumberLike {
+    Given: CucumberStepDefLike;
+    Then: CucumberStepDefLike;
+    When: CucumberStepDefLike;
+}
+
+interface CucumberStepDefLike {
+    (pattern: string, options: object, fn: () => void): void;
```
adamlacoste: This function definition is not perfectly accurate to how Cucumber actually works.  The second param, `options`, is optional but the TS build complains if we mark it as such.  Furthermore, the actual function signature of the `fn` step def handler (third param) allows for varying arguments and return type.  I'll refine this interface in future tickets if any of this gives me trouble.

modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
wild-love: Replied in the team chat.

modules/@specify-bdd/specify/src/lib/CucumberManager.ts
wild-love: How about `CucumberAPI`? Any object that wanted to be used here would need to do something like `class MyThing implements CucumberAPI { ...`

Or maybe `CucumberStepDefAPI`, or even `StepDefAPI` since step defs are cucumber-specific and dont need further specificity (maybe, Im not sure this is true in general, other tools might use this terminology).

modules/@specify-bdd/specify/src/lib/CucumberManager.ts
wild-love: You can mark an argument as being multiple potential types with a pipe "`|`". That'd look like 
`(pattern: string, secondArg: object | () => void): void;`

Not saying thats better than what you have, but might be an option. Or if cucumber exports a type to use here, that would work too. I think you could even do something like
```
export interface CucumberLike {
    Given: typeof Cucumber.Given;
    Then: typeof Cucumber.When;
    When: typeof Cucumber.Then;
}

File: modules/@specify-bdd/specify/src/lib/CucumberManager.ts (hunk 4 above)
Comment (wild-love) [reply modules/@specify-bdd/specify/src/lib/CucumberManager.ts:37]: Makes sense to me that `CucumberManager` would default to using brand-name `Cucumber`

File: modules/@specify-bdd/specify/src/lib/CucumberManager.ts (hunk 2 above)
Comment (adamlacoste) [reply modules/@specify-bdd/specify/src/lib/CucumberManager.ts:3]: I will need to expand this interface later to include non-step-def things, so I don't want to name it in a way that is too limiting.  I'm also reluctant to use API because that name makes me think of some RESTful service I might interact with.  Even if technically accurate, I'd like to avoid a name that leads to potential misapprehension of its purpose.

While `CucumberLike` is crude, I think it does communicate the purpose: this is a thing that is Cucumber-like, without necessarily needing to BE Cucumber (useful for testing purposes).

File: modules/@specify-bdd/specify/src/lib/CucumberManager.ts (hunk 5 above)
Comment (adamlacoste) [reply modules/@specify-bdd/specify/src/lib/CucumberManager.ts]: If I used pipes, wouldn't I have to duplicate the `fn` type?  Since it can appear in the second _or_ third param position, I mean.

The type returned by the `Given`, `When`, and `Then` methods appears to be [IDefineStep](https://cucumber.github.io/cucumber-js/variables/(root).Given.html), but I can't find documentation detailing what that is exactly or how we could import it.  I worry that using Cucumber's types would paint us into a corner as we would be forced to fulfill all the requirements it thinks are needed, as opposed to just those we want to test.

File: modules/@specify-bdd/specify/src/lib/CucumberManager.ts (hunk 2 above)
Comment (wild-love) [reply modules/@specify-bdd/specify/src/lib/CucumberManager.ts:3]: Fair enough, I can see a potenial avenue for confusion with `CucumberAPI`. I agree that `CucumberLike` is crude, but given the constraints, Im struggling to come up with good alternatives. We can move forward with that and rename later if inspiration strikes.

File: modules/@specify-bdd/specify/src/lib/CucumberManager.ts (hunk 5 above)
Comment (wild-love) [reply modules/@specify-bdd/specify/src/lib/CucumberManager.ts]: Ah, yeah. If we're matching cucumber's function signature, then we'd have to match the wonkiness, which make my suggestion a non-starter.

To your second point, it depends; If we are providing an abstracted/improved interface to the cucumber methods, then we wouldnt want to use Cucumber's `IDefineStep`. Alternatively, if we're only wrapping the cucumber function and maintaining the same input params, then it makes sense to reuse Cucumber's implementation typing to ensure our implementation never de-syncs.

File: modules/@specify-bdd/specify/src/lib/CucumberManager.ts [hunk 6]
```diff
@@ -0,0 +1,42 @@
+import Cucumber from "@cucumber/cucumber";
+
+export interface CucumberLike {
+    Given: typeof Cucumber.Given;
+    When: typeof Cucumber.When;
+    Then: typeof Cucumber.Then;
```
adamlacoste, line 6: In the end this seemed like the best way to get at these types without a lot of fudgy recreation of Cucumber code on our side.  `IDefineStep` isn't exported by Cucumber, nor are most of the other interfaces used in step definition.

wild-love (approved):

PR #235, opened 2025-12-09
Alternate shell types

wild-love:


features/plugin-cli/steps/shell.feature
```diff
@@ -27,6 +27,11 @@ Feature: Shell Step Definitions
         #     When the user kills another CLI shell
         #     Then there should be 0 active CLI shells
 
+        Scenario: Create alternative shell type
+            Given a "bash" shell
+            When the user runs the command "echo $0"
```
wild-love, line 32: Neat and tidier form of the shell type one-liner we've been discussing.

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
```diff
@@ -188,17 +209,12 @@ function startNamedDefaultShell(name?: string): void {
...
     this.cli.manager ??= new SessionManager();
     this.cli.manager.addSession(shell, name, this.fs.cwd);
-}
 
-/**
- * Start a default shell without a name.
- */
-function startDefaultShell(): void {
-    startNamedDefaultShell.call(this);
+    assert.ok(await this.cli.manager.validateShell(type), new AssertionError({ "message": "Shell validation failed." }));
```
wild-love: Somewhat vague error message, open to suggestions.

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
```diff
@@ -15,6 +15,8 @@ import type { SpawnOptions } from "node:child_process";
 
 Given("a/another CLI shell", startDefaultShell);
 
+Given("a {string} shell", startAltShell);
```
adamlacoste: We need to be consistent in our language and use "CLI shell" here and in the "when" version of this step below.

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
```diff
@@ -168,11 +172,28 @@ async function sendKillSignal(signal: string): Promise<void> {
 }
 
 /**
- * Start a default shell.
+ * Start a user-specified shell.
+ * 
+ * @param type - The type of shell to spawn (`sh`, `bash`, etc.)
+ */
+async function startAltShell(type: string): Promise<void> {
```
adamlacoste: I'm not wild about `type` being the way we describe which shell we use in code, since it has such a strong association with data types (particularly in TS) and may end up having a name collision with `just-type` or whatever other external library we bring in to manage runtime type validation.  Could we use something like `shellType` instead?  I'm not super particular about the name and open to other ideas, I just feel like `type` is misleading.

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
adamlacoste: I would make it something like `Failed to start ${type} CLI shell.` (with the caveat that I'd prefer not to use the varname `type` here either).

adamlacoste (changes requested):
Had a few language quibbles but functionally it looks good 👍

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
wild-love: Well of course you're not Wild 🧐

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
```diff
@@ -27,6 +29,8 @@ When("a/the user starts a/another CLI shell", startDefaultShell);
 
 When("a/the user starts a/an/the (async )command/process {refstr}", execCommand);
 
+When("a/the user starts a {string} CLI shell", startAltShell);
```
adamlacoste: Sorry for not spotting this before, but this should be `a/an {string} CLI shell` to account for the possibility that they're using something like `ash`.

modules/@specify-bdd/plugin-cli/src/lib/SessionManager.ts
```diff
@@ -329,6 +329,23 @@ export class SessionManager {
...
     }
 
+    /**
+     * Validate the shell type of the active session.
+     * 
+     * @param type - The expected shell type
+     * 
+     * @returns Whether its the correct type or not
+     */
+    async validateShell(type: string): Promise<boolean> {
```
adamlacoste: Missed a `type` that should be `shellType` here.

adamlacoste (changes requested):

adamlacoste (approved):

PR #236, opened 2025-12-09
defineStep method

### LINKED ISSUE #227: Add defineStep method
Add a new method to the `CucumberManager` class called `defineStep` which registers one or more enhanced Cucumber expressions (string or string[]), a handler function, and an optional options object with the managed Cucumber instance.

### LINKED ISSUE #230: Register multiple step expressions at once
When a user provides multiple Cucumber expressions via array as the first input parameter to `defineStep`, the method should register each one with Cucumber utilizing the method indicated by the (stripped off) first word of the expression.  (It makes no practical difference which method we use because they're all aliases under the hood, but we might as well be precise in case that changes in some future version of Cucumber.)

adamlacoste:
[#227] Added defineStep method (and associated unit test) to CucumberManager class

### DIFF
--- a/modules/@specify-bdd/specify/src/lib/CucumberManager.ts
+++ b/modules/@specify-bdd/specify/src/lib/CucumberManager.ts
@@ -46,7 +46,7 @@ export class CucumberManager {
      */
     defineStep(
         pattern: Array<StepDefPattern> | StepDefPattern,
-        handler: Function,
+        handler: () => void,
         options: StepDefOptions = {},
     ): CucumberManager {
         const patternList = Array.isArray(pattern) ? pattern : [pattern];
--- a/modules/@specify-bdd/specify/src/lib/CucumberManager.ts
+++ b/modules/@specify-bdd/specify/src/lib/CucumberManager.ts
@@ -13,7 +13,7 @@ interface StepDefHandlerFunction {
 }
 
 interface StepDefRegisterFunction {
-    (pattern: StepDefPattern, options?: StepDefOptions, fn?: StepDefHandlerFunction): void;
+    (pattern: StepDefPattern, options: StepDefOptions | StepDefHandlerFunction, fn?: StepDefHandlerFunction): void;
 }
 
 interface StepDefOptions {
--- a/modules/@specify-bdd/specify/src/lib/CucumberManager.ts
+++ b/modules/@specify-bdd/specify/src/lib/CucumberManager.ts
@@ -7,10 +7,6 @@ export interface CucumberLike {
     Then: typeof Cucumber.Then;
 }
 
-interface StepDefHandlerFunction {
-    (): void;
-}
-
 interface StepDefOptions {
     timeout?: number;
 }
@@ -50,7 +46,7 @@ export class CucumberManager {
      */
     defineStep(
         pattern: Array<StepDefPattern> | StepDefPattern,
-        handler: StepDefHandlerFunction,
+        handler: Function,
         options: StepDefOptions = {},
     ): CucumberManager {
         const patternList = Array.isArray(pattern) ? pattern : [pattern];
--- a/modules/@specify-bdd/specify/src/lib/CucumberManager.ts
+++ b/modules/@specify-bdd/specify/src/lib/CucumberManager.ts
@@ -2,7 +2,6 @@ import assert   from "node:assert";
 import Cucumber from "@cucumber/cucumber";
 
 export interface CucumberLike {
-    defineStep: typeof Cucumber.defineStep;
     Given: typeof Cucumber.Given;
     When: typeof Cucumber.When;
     Then: typeof Cucumber.Then;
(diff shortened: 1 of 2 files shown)


modules/@specify-bdd/specify/src/lib/CucumberManager.ts
```diff
@@ -1,3 +1,4 @@
+import assert   from "node:assert";
```
adamlacoste, line 1: We've talked about moving to Chai but I thought it was best not to bloat scope here.  I should make a ticket for that soon though.

modules/@specify-bdd/specify/src/lib/CucumberManager.ts
```diff
@@ -27,6 +38,61 @@ export class CucumberManager {
...
+                const raw   = patternList[i].toString();
+                const match = raw.match(/^\/\^?(Given|When|Then) /);
+                assert(match, `Invalid pattern expression: ${raw}`);
+
+                const endIndex = raw.lastIndexOf("/");
+                const trimmed  = raw.slice(match[0].length, endIndex);
+
+                patternKey = match[1];
+                patternParsed = new RegExp(trimmed, raw.slice(endIndex + 1));
+            }
```
adamlacoste, line 74: This conditional logic might later move into a `parsePattern` method or something like that.  It seemed fine to me at this level of complexity but when I add enhanced pattern pre-processing, it's gonna blow up.  I opted to leave the optimization for when it's warranted though.

modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
```diff
@@ -2,29 +2,149 @@ import { CucumberManager } from "../CucumberManager";
...
+    "When":  fakeDefineStep,
+};
 
 describe("CucumberManager", () => {
-    describe("cucumber property", () => {
-        it("returns a Cucumber-like object", () => {
-            const cm = new CucumberManager(cucumber);
+    describe("properties", () => {
+        describe("cucumber", () => {
+            it("returns a Cucumber-like object", () => {
```
wild-love: Small nitpick; I might modify the phrasing here, since its not just any ol' cucumber object we get back, its specifically the one given to the constructor.

modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
```diff
@@ -2,29 +2,149 @@ import { CucumberManager } from "../CucumberManager";
...
+                it("a single regular expression", () => {
+                    const noKeyPatStr = "I do something with .*";
+                    const noKeyPatReg = new RegExp(noKeyPatStr, "i");
+                    const keyPatStr   = `When ${noKeyPatStr}`;
+                    const keyPatReg   = new RegExp(keyPatStr, "i");
+
+                    cm.defineStep(keyPatReg, fakeHandler);
+
+                    expect(cm.cucumber.When).toHaveBeenCalledTimes(1);
+                    expect(cm.cucumber.When).toHaveBeenCalledWith(noKeyPatReg, {}, fakeHandler);
```
wild-love: This is a bit ambiguous, we dont know what this value is because we never see it directly in the test (its calculated on line 55). I think itd be more clear if we're explicit about the regex we want to see here, using a typed out regex.

modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
```diff
@@ -2,29 +2,149 @@ import { CucumberManager } from "../CucumberManager";
...
+                });
+
+                it("multiple mixed expressions", () => {
+                    const noKeyPatStr = "I do something with {param}";
+                    const noKeyPatReg = new RegExp(noKeyPatStr.replace("{param}", ".*"));
```
wild-love: Same thing here as above, these might be a bit too fancy 😅

modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
```diff
@@ -2,29 +2,149 @@ import { CucumberManager } from "../CucumberManager";
...
+
+                        expect(() => cm.defineStep(pat, fakeHandler)).toThrow(
+                            `Invalid pattern expression: ${pat}`,
+                        );
+                    });
+
+                    it("improper formatting", () => {
+                        const pat = " When I do something with {param}";
+
+                        expect(() => cm.defineStep(pat, fakeHandler));
```
wild-love: The actual test here seems to be missing 🤔

modules/@specify-bdd/specify/src/lib/CucumberManager.ts
```diff
@@ -6,6 +7,16 @@ export interface CucumberLike {
     Then: typeof Cucumber.Then;
 }
 
+interface StepDefHandlerFunction {
+    (): void;
```
wild-love: What the heck is this!? Never seen this one before 🧐

wild-love (changes requested):

modules/@specify-bdd/specify/src/lib/CucumberManager.ts
adamlacoste: Unnecessary chaff, since this interface is used in literally just one place; I'll use an inline function type description instead.

wild-love (approved):

PR #238, opened 2025-12-11
Dev/86/named shells

wild-love:


modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
```diff
@@ -15,8 +15,12 @@ import type { SpawnOptions } from "node:child_process";
 
 Given("a/another CLI shell", startDefaultShell);
 
+Given("a/another CLI shell named {string}", startDefaultNamedShell)
+
 Given("a/an {string} CLI shell", startAltShell);
 
+Given("a/an {string} CLI shell named {string}", startAltNamedShell);
```
wild-love, line 22: Probably a topic for a meeting, but I'd like to suggest that we change the above (and all other instances of usage) to make "CLI" optional:

`Given("a/an {string} (CLI )shell named {string}", startAltNamedShell);`

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
```diff
@@ -223,14 +254,46 @@ async function startShell(shellType: string = "sh", name?: string): Promise<void
  * @throws AssertionError
  * If there is no SessionManager initialized.
  */
-function switchShell(): void {
+function cycleShell(): void {
```
wild-love: Inspiration struck to rename this to be more pointed that we arent selecting what we're switching to. Happy to change it back if there are objections.

modules/@specify-bdd/plugin-cli/src/lib/tests/SessionManager.test.ts
```diff
@@ -497,6 +497,56 @@ describe("SessionManager", () => {
...
+            it("switches to the indexed session", () => {
+                expect(sessionManager.activeSession.session).toBe(altSession2);
+
+                sessionManager.switchToSession(0);
+
+                expect(sessionManager.activeSession.session).toBe(altSession1);
+
+                sessionManager.switchToSession(2);
+
+                expect(sessionManager.activeSession.session).toBe(altSession3);
```
wild-love, line 536: I dont prefer alternating state changes and assertions like this (and in the above), but its the only way to be 100% the functions is doing what its supposed to 🤷

modules/@specify-bdd/plugin-cli/src/lib/SessionManager.ts
```diff
@@ -329,6 +329,26 @@ export class SessionManager {
...
 
+    /**
+     * Switch to the session that matches the given selector.
+     * 
+     * @param selector - The index or name of the session to switch to
+     *
+     * @throws AssertionError
+     * If there is no matching session
+     */
+    switchToSession(selector: number | string): void {
```
wild-love, line 340: Im on the fence about whether I like this multi-type argument. On the one hand its pretty convenient and helps keep the class a little more trim, but it also irks me a bit. I think its right on the line of when I would typically split off into a new function. Maybe we keep it this way for now, and split it up if the function becomes any more complex.

modules/@specify-bdd/plugin-cli/src/lib/SessionManager.ts
```diff
@@ -329,6 +329,26 @@ export class SessionManager {
...
+        const session = typeof selector === "string"
+            ? this.#sessions.find((session) => session.name === selector)
+            : this.#sessions[selector];
+        
+        if (!session) {
+            throw new Error(`No session found with selector: ${selector}`);
```
wild-love: Selector might not be the best name, but its the best I could think of 😅 Open to suggestions!

features/plugin-cli/steps/shell.feature
```diff
@@ -31,6 +31,13 @@ Feature: Shell Step Definitions
             Given a "bash" CLI shell
             When the user runs the command "echo $0"
             Then the last command's terminal output should match "^bash$"
+        
+        Scenario: Create alternative shell type with a name
+            Given a "bash" CLI shell named "test-shell"
+            And another CLI shell
+            When a user switches to the CLI shell named "test-shell"
+            And the user runs the command "echo $0"
+            Then the last command's terminal output should match "^bash$"
```
adamlacoste, line 40: I would prefer if we had a scenario for creating a default named shell as well as an alternative named shell.  I realize that under the hood they're all using the same logic, but covering all our bases and demonstrating the functionality via our tests are both important priorities for these specs.

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
adamlacoste: I would be OK with making "CLI" optional in all our step definitions that reference "CLI shell".  But maybe that deserves its own ticket?

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
```diff
@@ -27,11 +31,19 @@ When("a/the user sends a {cliSignal} signal to the last command", sendKillSignal
...
+When("a/the user starts a/another CLI shell named {string}", startDefaultNamedShell);
+
 When("a/the user starts a/an/the (async )command/process {refstr}", execCommand);
 
 When("a/the user starts a/an {string} CLI shell", startAltShell);
 
-When("a/the user switches shells", switchShell);
+When("a/the user starts a/an {string} CLI shell named {string}", startAltNamedShell);
+
+When("a/the user switches shells", cycleShell);
```
adamlacoste: For now, let's add "CLI" to this one (and any scenarios that reference it) so we're being consistent.  If we change the convention to make "CLI" optional later, it will be easier to do a global find and replace if these steps all use identical phrasing.

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
adamlacoste: I can understand the desire, but the step definition this handler serves uses the "switch" language, so IMO it should be consistent.

I think it might be worth merging this handler with that of `selectShellByIndex`.  Personally, I'd keep the name `switchShell` and accept one param that can be either a number, a string, or nothing (for the "cycle" version).  Since all three functions are identical but for one line, it would be simple to do some kind of `switch (typeof target)` structure or something like that.  Feel free to keep them separate if you prefer it, but to me it seems like we're repeating boilerplate by keeping them separate.

modules/@specify-bdd/plugin-cli/src/lib/tests/SessionManager.test.ts
adamlacoste: Personally I think the switch-to-and-verify-session3 step in both the above tests is extraneous, but overtesting is better than undertesting IMO so I don't object to keeping it.  I wouldn't have mentioned it at all except to point out that if you removed it, you would have less action/assertion alternation. 🤷

modules/@specify-bdd/plugin-cli/src/lib/tests/SessionManager.test.ts
```diff
@@ -497,6 +497,56 @@ describe("SessionManager", () => {
...
+            });
+
+            it("throws if the session does not exist", () => {
+                expect(() => sessionManager.switchToSession("bad-name")).toThrow(
+                    "No session found with selector: bad-name",
+                );
+
+                expect(() => sessionManager.switchToSession(-1)).toThrow(
+                    "No session found with selector: -1",
+                );
```
adamlacoste, line 546: This test is perfectly fine for now as an accurate boundary for what the expected functionality should be at this stage, but it does make me think that we might want to make a ticket to support negative indices at some point in the future. 🤔

modules/@specify-bdd/plugin-cli/src/lib/SessionManager.ts
adamlacoste: 🤷 I like it.  The action being described here is highly specific and targeted: switching to another session.  Only the means we use to identify that session varies, and ultimately the logic for supporting that variation is pretty simple.

modules/@specify-bdd/plugin-cli/src/lib/SessionManager.ts
adamlacoste: Maybe `identifier`?  But I think it would be better to have different messages depending on what type of selector was provided, rather than one message that works for all.

adamlacoste (changes requested):

features/plugin-cli/steps/shell.feature
wild-love: - [x] Implement new scenario

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
wild-love: I agree, that work should be its own ticket.

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
wild-love: - [x] Fix inconsistency

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
wild-love: Cucumber seems to have an issue with optional params, it doesnt like it when the arguments arent set in stone. Thats the main reason I split these functions out. If there were a way to make cucumber be satisfied with optional arguments then I'd be happy to merge all of these.

wild-love: - [x] Revert change

modules/@specify-bdd/plugin-cli/src/lib/SessionManager.ts
wild-love: - [x] Make messaging type-responsive

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
adamlacoste: Oh yeah, we will need three separate handler functions no matter what.  I was suggesting we might want to merge the internal logic though.  Three handlers would each call a utility function that can accept a string, number, or nothing.  Sorry for burying the lede there.

wild-love: - [x] Dry up handler boilerplate duplication

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
```diff
@@ -286,10 +278,28 @@ function selectShellByIndex(index: number): void {
...
+ * If there is no SessionManager initialized.
+ */
+function switchShell(selector?: number | string): void {
     assert.ok(this.cli.manager, new AssertionError({ "message": "No shell session initialized." }));
 
-    this.cli.manager.switchToSession(name);
+    if (!selector) {
+        this.cli.manager.switchToNextSession();
+    }
+    else {
```
adamlacoste: Did you run lint on this?  I would have thought it would fix this.  Should be `} else {` on one line.

adamlacoste (approved):

adamlacoste (approved):

PR #239, opened 2025-12-19
Specify entrypoint

### LINKED ISSUE #226: Export defineStep function from Specify entrypoint

adamlacoste:


modules/@specify-bdd/specify/src/lib/tests/integration/index.test.ts
```diff
@@ -0,0 +1,23 @@
+import { defineStep } from "../../../index";
```
adamlacoste, line 1: This seems like a place where it would have been better to use a TS shortcut like `@/index`, but my attempts to do so resulted in vitest being unable to find the module.  Maybe I just don't understand how this feature works?  But it seemed notable to me that in src/exec.ts we're using `@/config` (meaning src/config) but also a bunch of `./lib/*` imports, so it would be nice if we could figure out how to use this feature consistently to avoid a lot of unnecessary tree traversal in our import paths.

modules/@specify-bdd/specify/src/lib/tests/integration/index.test.ts
```diff
@@ -0,0 +1,23 @@
...
+
+const { mockGiven } = vi.hoisted(() => {
+    return { "mockGiven": vi.fn() };
+});
+
+vi.mock("@cucumber/cucumber", () => {
+    const cucumber = { "Given": mockGiven };
+
+    return { "default": cucumber, ...cucumber };
+});
```
adamlacoste, line 11: This was (apparently?) all the mocking I had to do to get this int test to pass.  I don't understand it well enough to be confident that I'm using it right, so please double-check this section thoroughly.  The intention here was to mock the `@cucumber/cucumber` import that `CucumberManager` does, and for the mocked module to have just the export I need it to have, `Given`, in order for my test to pass.

modules/@specify-bdd/specify/src/lib/tests/integration/index.test.ts
```diff
@@ -0,0 +1,23 @@
...
+
+describe("entrypoint module (integration)", () => {
+    describe("exports", () => {
+        describe("defineStep()", () => {
+            it("registers a step definition with Cucumber", () => {
+                defineStep("Given a step definition", () => true);
+
+                expect(mockGiven).toHaveBeenCalled();
```
adamlacoste, line 19: I thought it best not to over-test here, because we don't want to redundantly verify all the same functionality as the `CucumberManager` test, we just want to make sure that the parts are linked up correctly and our entrypoint module's `defineStep` export, when called, properly uses `CucumberManager.defineStep` which in turn calls (mocked) Cucumber's `Given` function.

modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
adamlacoste, line 1: I noticed I wasn't being too specific with the types here, and since we have them, I might as well use them.

modules/@specify-bdd/specify/src/exec.ts
adamlacoste, line 1: The artist formerly known as main.ts will now be called exec.ts.

modules/@specify-bdd/specify/package.json
adamlacoste, line 1: Apparently `main` is how dinosaurs defined their package entrypoing back in the Cretaceous period but now we're supposed to use `exports`.  While you can define both, the only reason to do so is to maintain backwards compatibility with Node v10 or something like that, which we definitely don't support.

modules/@specify-bdd/specify/src/lib/CucumberManager.ts
adamlacoste, line 1: Previously, we had a dependency on Cucumber in here because we wanted to ensure any use of `CucumberManager.getInstance()` would use the same version of Cucumber.  However, the introduction of the index module as an entrypoint made this unnecessary, because that's now going to be the only place `CucumberManager.getInstance()` gets called.

I would have liked to remove the import of `@cucumber/cucumber` altogether, but because we're using it for type inference in our `CucumberLike` interface, I decided to instead change it to an `import type` which gets removed from the built Javascript, at least.  Now this module's dependency on Cucumber is purely for type-checking.

modules/@specify-bdd/specify/src/index.ts
```diff
@@ -0,0 +1,25 @@
+import Cucumber from "@cucumber/cucumber";
```
adamlacoste, line 1: The Cucumber dependency removed from `CucumberManager` moved here.

modules/@specify-bdd/specify/src/index.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { CucumberManager } from "./lib/CucumberManager";
+
+import type { StepDefOptions, StepDefPattern } from "./lib/CucumberManager";
+
+const cm = CucumberManager.getInstance(Cucumber);
```
adamlacoste, line 7: The entrypoint module handles the initialization of `CucumberManager` so all external usage should share the same instance of CM and Cucumber.  I believe that the way module imports work should guarantee that this module-scoped code defining the `cm` variable only fires once for any given Node.js runtime, but in the event it did execute more than once, the singleton design of `getInstance` ought to ensure consistency anyway.

modules/@specify-bdd/specify/src/index.ts
```diff
@@ -0,0 +1,25 @@
...
+ */
+export function defineStep(
+    pattern: Array<StepDefPattern> | StepDefPattern,
+    handler: () => void,
+    options: StepDefOptions = {},
+): void {
+    cm.defineStep(pattern, handler, options);
+}
+
+export default { defineStep };
```
adamlacoste, line 25: We are exporting both the function `defineStep` as a named export and an object with a `defineStep` method as the default.

modules/@specify-bdd/specify/src/lib/tests/integration/index.test.ts
wild-love: I agree, though if `@/index` doesnt work then Im not sure what the answer would be 🤷 

Edit: on second thought, this is a test file, which I dont think follows the same TS config rules as the rest of the application since its transpiled on demand by Vitest when the tests are run.

modules/@specify-bdd/specify/src/lib/tests/integration/index.test.ts
wild-love: If you dont need any particular mock implementation, then you could probably reduce the lines 3-11 to just `vi.mock("@cucumber/cucumber")`, which would auto-mock all properties of the specified module.

modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
```diff
@@ -1,8 +1,12 @@
 import { CucumberManager } from "../CucumberManager";
 
+import type { StepDefOptions, StepDefPattern } from "../CucumberManager";
+
 import type { CucumberLike } from "../CucumberManager";
 
-function fakeDefineStep(pattern, options = {}, handler = () => {}) {
+function fakeDefineStep(pattern: StepDefPattern, options: StepDefOptions = {}, handler = () => {}) {
+    // the following logic only exists to prevent our linter from complaining
+    // that the params above are defined but not used
```
wild-love: This is fine, though we could also do `// eslint-disable-next-line {name-of-broken-rule-here}` on line 6 if we didnt want code for the sake of linter appeasement

modules/@specify-bdd/specify/src/index.ts
```diff
@@ -0,0 +1,25 @@
+import Cucumber from "@cucumber/cucumber";
+
+import { CucumberManager } from "./lib/CucumberManager";
```
wild-love, line 3: Can we use `@/lib/CucumberManager` instead? Its functionally the same, but I think its best to remain consistent.

modules/@specify-bdd/specify/src/lib/tests/integration/index.test.ts
adamlacoste: That is likely so, but it seems to me that we should figure out how to make the test files work like everything else.  But that's out of scope for this ticket so we'll have to figure it out another time.

modules/@specify-bdd/specify/src/lib/tests/integration/index.test.ts
adamlacoste: - [x] Try Wild's Cucumber mock implementation suggestion

modules/@specify-bdd/specify/src/lib/tests/integration/index.test.ts
adamlacoste: Given the above comment about auto-mocking Cucumber, what would we `expect` to have been called here?  `cucumber.Given`?

modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
adamlacoste: - [x] Replace placeholder code in `fakeDefineStep` with eslint disable comment

modules/@specify-bdd/specify/src/index.ts
adamlacoste: - [x] Try making the `CucumberManager` import in Specify's index module use a `@` path shortcut

adamlacoste: These changes don't work when we run unit tests.

```
 FAIL  modules/@specify-bdd/specify/src/lib/tests/integration/index.test.ts [ modules/@specify-bdd/specify/src/lib/tests/integration/index.test.ts ]
Error: Cannot find package '@/lib/CucumberManager' imported from '/Users/adam/Development/specify-bdd/core/modules/@specify-bdd/specify/src/index.ts'
 ❯ modules/@specify-bdd/specify/src/index.ts:3:1
      1| import Cucumber from "@cucumber/cucumber";
      2|
      3| import { CucumberManager } from "@/lib/CucumberManager";
       | ^
      4|
      5| import type { StepDefOptions, StepDefPattern } from "@/lib/CucumberManager";

Caused by: Error: Failed to load url @/lib/CucumberManager (resolved id: @/lib/CucumberManager) in /Users/adam/Development/specify-bdd/core/modules/@specify-bdd/specify/src/index.ts. Does the file exist?
 ❯ loadAndTransform node_modules/.pnpm/vite@7.0.0_@types+node@22.15.18_jiti@2.4.2_tsx@4.19.4_yaml@2.8.0/node_modules/vite/dist/node/chunks/dep-Bsx9IwL8.js:26427:33
```

I definitely think it SHOULD work and I would like to make it so at some point.

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
adamlacoste, line 1: These changes (and all the others in `plugin-cli`) were made by the linter; it seems that some unlinted changes snuck into a previous merge to `develop`.  I suspect this happened because the automated PR checks just verify that `pnpm lint` doesn't produce any fatal errors, it doesn't attempt a `lint:fix` and then verify that no changes were made.  We might want to revisit that automation at some point in the future so that linter fixes like this happen as part of the PR they really belong to.

modules/@specify-bdd/specify/src/lib/tests/integration/index.test.ts
adamlacoste: We tried an auto-mock and ran into an error about "context" that was difficult to understand.  There are probably complexities to mocking Cucumber which we don't understand yet.  For now, the limited mock I created here will serve the purpose we need it to just fine.

modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
```diff
@@ -1,16 +1,15 @@
...
 import type { CucumberLike } from "../CucumberManager";
 
-function fakeDefineStep(pattern, options = {}, handler = () => {}) {
-    if (pattern && options) {
-        handler();
-    }
-}
+// prettier-ignore
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+function fakeDefineStep(pattern: StepDefPattern, options: StepDefOptions = {}, handler = () => {}) {}
```
adamlacoste, line 9: This method's placeholder logic (and that of the one below) has been removed and linter disable comments are in place to suppress complaints about it.  The reason this one needs two separate disable comments (for prettier and eslint both) is that prettier wants to reformat this function definition across 5 lines (including one for each param), which made the single-line eslint disable comment apply only to the first of those lines.  While I could have opted for a block disable, I thought it was unnecessary to spread this placeholder function definition across so much white space so I opted for the compact solution.

modules/@specify-bdd/specify/src/index.ts
adamlacoste: We opted to stick with the relative paths for now; this issue should be fixed in the next PR.

adamlacoste (commented):
All feedback addressed.

modules/@specify-bdd/specify/src/lib/tests/integration/index.test.ts
wild-love: In our meeting today, we discovered the next PR addresses exactly this issue 😂

wild-love (approved):

wild-love (approved):

PR #240, opened 2025-12-23
Enhanced syntax preprocessing

### LINKED ISSUE #231: Preprocess enhanced syntax in Cucumber expressions
Whenever a step definition expression includes square brackets (`[ ]`), we should interpret it as a Specify-enhanced pattern and preprocess the contents of those brackets accordingly.  The following rules govern how we parse enhanced expression syntax.

1. Split the substring on any forward slash (`/`) character and create variants of the whole expression which replace the enhanced clause with each split piece.
2. If the words "I" or "the user" appear in the enhanced clause and are not already optional, make them so using parentheses (`( )`).

All variants of a given expression generated by this preprocessing logic should subsequently be registered with Cucumber.

adamlacoste:
[#231] Implemented enhanced syntax preprocessing and unit tests verifying its behavior

This PR incorporates (and merges into) #239 so that should be reviewed & merged first.


modules/@specify-bdd/specify/src/lib/CucumberManager.ts
```diff
@@ -52,43 +57,123 @@ export class CucumberManager {
...
-                const raw   = patternList[i].toString();
-                const match = raw.match(/^\/\^?(Given|When|Then) /);
-                assert(match, `Invalid pattern expression: ${raw}`);
-
-                const endIndex = raw.lastIndexOf("/");
-                const trimmed  = raw.slice(match[0].length, endIndex);
-
-                patternKey = match[1];
-                patternParsed = new RegExp(trimmed, raw.slice(endIndex + 1));
+                variants.push(...this.#parseRegularExpressionVariants(patternList[i] as RegExp));
```
adamlacoste, line 76: This will always add just one element to the variants array, but it seemed best to make the handling of string and regular expressions as consistent as possible.

modules/@specify-bdd/specify/src/lib/CucumberManager.ts
```diff
@@ -52,43 +57,123 @@ export class CucumberManager {
...
+                        this.cucumber.Given(variant.pattern, options, handler);
+                        break;
+                    case "When":
+                        this.cucumber.When(variant.pattern, options, handler);
+                        break;
+                    case "Then":
+                        this.cucumber.Then(variant.pattern, options, handler);
+                        break;
+                    default:
+                        assert.fail(`Invalid pattern keyword: ${variant.keyword}.`);
```
adamlacoste, line 91: I don't think there's a way for this default case to fire, but I'd rather have it and not need it than the other way around.

modules/@specify-bdd/specify/src/config/content.config.ts
```diff
@@ -31,6 +31,9 @@ const contentObj = {
             },
         },
     },
+    "specifications": {
```
wild-love, line 34: What is the context for the term "specifications"?

modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
```diff
@@ -1,22 +1,29 @@
...
+}
 
 const cucumber: CucumberLike = {
     "Given": fakeDefineStep,
     "Then":  fakeDefineStep,
     "When":  fakeDefineStep,
 };
 
+const cmOpts = {
+    "subjects": ["I", "the user"],
```
wild-love, line 19: Is there a way to import the config props for this, so we have a single source of truth for subjects?

modules/@specify-bdd/specify/src/lib/CucumberManager.ts
```diff
@@ -53,53 +68,132 @@ export class CucumberManager {
...
+        }
+
+        let variants = [subphrase];
+
+        // next, process alternates (/)
+        variants = variants.flatMap((variant) => variant.split("/"));
+
+        // then, look for known subjects and make them optional
+        variants = variants.flatMap((variant) => {
+            // TODO: what should we do to make these subjects more manageable/configurable?
```
wild-love: Good question 🤔

modules/@specify-bdd/specify/src/lib/CucumberManager.ts
```diff
@@ -53,53 +68,132 @@ export class CucumberManager {
...
+            return variant;
+        });
+
+        // finally, substitute every variant back into the original phrase
+        return variants.map((variant) => phrase.replace(enhanced, variant));
+    }
+
+    /**
+     * Parse a regular expression and return a list of pattern variants to feed
+     * into Cucumber.  (This will always be a list with one item, but doing this
```
wild-love, line 163: Might be good to use the following format here (or similar): 
```/**
     * Parse a regular expression and return a list of pattern variants to feed
     * into Cucumber. 
     * 
     * @note
     * This will always be a list with one item, but doing this keeps
     * consistency with the enhanced expression parser.

File: modules/@specify-bdd/specify/src/lib/CucumberManager.ts [hunk 7]
```diff
@@ -53,53 +68,132 @@ export class CucumberManager {
...
+
     /**
      * Get the singleton instance of CucumberManager.
      *
      * @param cucumber - The Cucumber object to manage
      *
      * @returns The CucumberManager instance
      */
-    static getInstance(cucumber: CucumberLike): CucumberManager {
+    static getInstance(cucumber: CucumberLike, options: ManagerOptions): CucumberManager {
```
wild-love: Shouldnt these be optional params, since theyre unused when retrieving the instance?

Alternatively, maybe instead of `getInstance()` handling this, we could have an `initialize()` or something that does nothing (or errors) when run when the instance exists

modules/@specify-bdd/specify/src/config/content.config.ts
adamlacoste: I meant for it to reference behavior specifications, i.e. Gherkin feature docs.  I'm open to other terminology.

modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
adamlacoste: I'm sure there is, but I'm not sure we _should_.  In most other tests we keep test data in the test module where it will be used, so that if a config change is made, the test will fail until the test is updated.

modules/@specify-bdd/specify/src/lib/CucumberManager.ts
adamlacoste: Oh, I think I need to remove this comment.  It's a relic of an earlier iteration of this, when the subjects were hard-coded.

- [x] Remove comment about subject configurability

modules/@specify-bdd/specify/src/lib/CucumberManager.ts
adamlacoste: - [x] Add a note to the `#parseRegularExpressionVariants` docblock

modules/@specify-bdd/specify/src/lib/CucumberManager.ts
adamlacoste: I'm OK with making the params optional, that makes sense to me.

I'm not sure about having an `initialize` static method, which would presumably do only what the conditional does now in approximately the same number of lines.  I don't think it's necessary to throw an error; remember that this `getInstance` method is only called in our index module and is not exposed outside of the core `specify` package.

- [x] Make `getInstance` method params optional

modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
wild-love: Ah, didnt realize this was a test file. Carry on 😅

wild-love (approved):

PR #243, opened 2026-01-15
Duplicate expressions

### LINKED ISSUE #232: Screen step expressions for duplicates
As step definition expressions are being (pre)processed and registered, we should attempt to extrapolate all the permutations which may result from a given expression's optional and alternate syntax, including variants produced via enhanced syntax, and store them all for later reference, along with the call stack that invoked `defineStep`.  Prior to registering any expression with Cucumber, we should check it against the permutation store to verify that the same pattern hasn't already been registered, because distinct Cucumber expressions can still produce identical signatures in some cases.  If a match is found in the permutation store, the `defineStep` method should throw an error explaining the duplication issue and clearly identifying the call stacks of the two duplicate permutations.  This will serve as an early warning system for users who have inadvertently created ambiguous step definitions.

adamlacoste:
Resolves #232.

### DIFF
--- a/modules/@specify-bdd/specify/src/cucumber/support/hooks.ts
+++ b/modules/@specify-bdd/specify/src/cucumber/support/hooks.ts
@@ -8,8 +8,10 @@ import { After, Before, BeforeAll } from "@cucumber/cucumber";
 import { globby                   } from "globby";
 import path                         from "node:path";
 import { pathToFileURL            } from "node:url";
+import { error                    } from "node:console";
 
-import { config } from "@/config/all";
+import { config          } from "@/config/all";
+import { CucumberManager } from "@/lib/CucumberManager";
 
 import type { JsonObject } from "type-fest";
 
@@ -31,6 +33,16 @@ BeforeAll(async function () {
             }).then((mod) => ({ "ref": mod.default }) as JsonObject),
         ),
     );
+
+    // warn if CucumberManager registered any duplicate step patterns
+    // but only do it once per run (even if running multiple parallel workers)
+    if (!process.env.CUCUMBER_WORKER_ID || process.env.CUCUMBER_WORKER_ID === "0") {
+        Object.entries(CucumberManager.getInstance().patterns)
+            .filter((entry) => entry[1] > 1)
+            .forEach((entry) =>
+                error("WARNING: pattern '%s' was registered %d times.", entry[0], entry[1]),
+            );
+    }
 });
 
 Before({ "name": "Core before hook" }, async function (data) {
--- a/modules/@specify-bdd/specify/src/index.ts
+++ b/modules/@specify-bdd/specify/src/index.ts
@@ -1,4 +1,4 @@
-import Cucumber from "@cucumber/cucumber";
+import * as Cucumber from "@cucumber/cucumber";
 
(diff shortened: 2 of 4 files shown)


modules/@specify-bdd/specify/src/cucumber/support/hooks.ts
```diff
@@ -8,8 +8,10 @@ import { After, Before, BeforeAll } from "@cucumber/cucumber";
 import { globby                   } from "globby";
 import path                         from "node:path";
 import { pathToFileURL            } from "node:url";
+import { error                    } from "node:console";
```
adamlacoste, line 11: I still think it's real dumb that we have to do this.  The purpose, AFAIK, of having lint rules forbidding `console.error` is to force ourselves to remove any such actions.  The rationale for that, I presume, is that in front-end JS development, console output is a debugging tool that should be excised from production code.  If that rationale was worth upholding, then we shouldn't allow ourselves the workaround of deconstructing `console` to get at its output methods.

But we're not writing front-end code and writing to the console is, at least some of the time, an intentional way we communicate information to our users.  So IMO the linter rule forbidding `console`'s use is misguided and we should remove it.  Then it would no longer be necessary to do this import deconstruction for a utility that is already globally available.

modules/@specify-bdd/specify/src/lib/CucumberManager.ts
```diff
@@ -77,18 +83,25 @@ export class CucumberManager {
             }
 
             for (const variant of variants) {
-                switch (variant.keyword) {
+                const key = variant.keyword;
+                const pat = variant.pattern;
```
adamlacoste, line 87: I was repeating these `variant.` references so frequently that I figured it would be tidier to shorten them.  I think the repetition was more egregious in an earlier iteration of this work when I was throwing an error in here though, and if you wanted this changed back to save on unnecessary variables, I think it would be perfectly fine the way it was before.

modules/@specify-bdd/specify/src/index.ts
```diff
@@ -1,4 +1,4 @@
-import Cucumber from "@cucumber/cucumber";
+import * as Cucumber from "@cucumber/cucumber";
```
adamlacoste, line 1: Interesting bug I only discovered while trying to test this: `Cucumber` doesn't have a default export so I had to bundle all the named exports to make one.

modules/@specify-bdd/specify/src/cucumber/support/hooks.ts
wild-love: You mention it above, but it seems to me like this is doing the job it was meant to do. By making the conscious decision to import, we ensure no debugging/unintentional output was left in, regardless of frontend/cli distinction. We would move to a logger of some sort, but we'd still have to import it... 🤔

modules/@specify-bdd/specify/src/lib/CucumberManager.ts
wild-love: I think the temp variables are fine. My quibble would be the shortening of `pattern` to `pat`, and the same for `keyword`, but its not a big deal 🙃

wild-love (approved):

PR #245, opened 2026-01-16
defineParamType and defineWorld

### LINKED ISSUE #228: Add defineParam method
Add a method to `CucumberManager` called `defineParam` which serves as a wrapper for Cucumber's `defineParameterType`.

### LINKED ISSUE #229: Add setWorld method
Add a method to `CucumberManager` called `setWorld` which serves as a wrapper for Cucumber's `setWorldConstructor`.

adamlacoste:
Resolves #228 and #229.

### DIFF
--- a/modules/@specify-bdd/specify/src/index.ts
+++ b/modules/@specify-bdd/specify/src/index.ts
@@ -40,9 +40,11 @@ export function defineStep(
 }
 
 /**
- * Register a custom world constructor with Specify.  It is STRONGLY recommended
- * that your custom world extends the default Specify world and calls super() in
- * its constructor.
+ * Register a custom world constructor with Specify.
+ *
+ * @remarks
+ * It is STRONGLY recommended that your custom world extends the default Specify
+ * world and calls super() in its constructor.
  *
  * @param world - A custom world
  */
--- a/modules/@specify-bdd/specify/src/index.ts
+++ b/modules/@specify-bdd/specify/src/index.ts
@@ -3,12 +3,22 @@ import Cucumber from "@cucumber/cucumber";
 import { config          } from "@/config/all";
 import { CucumberManager } from "@/lib/CucumberManager";
 
-import type { ParamTypeOptions, StepDefOptions, StepDefPattern } from "@/lib/CucumberManager";
+import type {
+    ParamTypeOptions,
+    StepDefOptions,
+    StepDefPattern,
+    WorldLike,
+} from "@/lib/CucumberManager";
 
 const cm = CucumberManager.getInstance(Cucumber, {
     "subjects": config.content.specifications.subjects,
 });
 
+/**
+ * Register a new parameter type with Specify.
+ *
+ * @param options - The param options
+ */
 export function defineParamType(options: ParamTypeOptions): void {
     cm.defineParamType(options);
 }
@@ -29,4 +39,15 @@ export function defineStep(
     cm.defineStep(pattern, handler, options);
 }
 
-export default { defineParamType, defineStep };
+/**
+ * Register a custom world constructor with Specify.  It is STRONGLY recommended
+ * that your custom world extends the default Specify world and calls super() in
+ * its constructor.
+ *
+ * @param world - A custom world
+ */
+export function defineWorld(world: WorldLike): void {
+    cm.defineWorld(world);
+}
+
+export default { defineParamType, defineStep, defineWorld };
--- a/modules/@specify-bdd/specify/src/index.ts
+++ b/modules/@specify-bdd/specify/src/index.ts
@@ -3,11 +3,15 @@ import Cucumber from "@cucumber/cucumber";
 import { config          } from "@/config/all";
 import { CucumberManager } from "@/lib/CucumberManager";
 
-import type { StepDefOptions, StepDefPattern } from "@/lib/CucumberManager";
+import type { ParamTypeOptions, StepDefOptions, StepDefPattern } from "@/lib/CucumberManager";
 
 const cm = CucumberManager.getInstance(Cucumber, {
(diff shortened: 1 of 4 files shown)


modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
```diff
@@ -1,18 +1,33 @@
...
-import type { CucumberLike, StepDefOptions, StepDefPattern } from "@/lib/CucumberManager";
+import type {
+    CucumberLike,
+    ParamTypeOptions,
+    StepDefOptions,
+    StepDefPattern,
+    WorldLike,
+} from "@/lib/CucumberManager";
+
+/* eslint-disable @typescript-eslint/no-explicit-any, @typescript-eslint/no-unused-vars */
```
adamlacoste, line 11: Because I had to define more dummy functions which didn't use their params, I moved this ESLint disable over to block format.  I decided to drop the prettier ignore line because now that the ESLint exception is a block, it isn't critical to keep the dummy function definitions on one line.  I also added an exception to the prohibition on `any` since user-defined functions could theoretically return literally anything.

modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
```diff
@@ -1,18 +1,33 @@
...
 
-// prettier-ignore
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-function fakeDefineStep(pattern: StepDefPattern, options: StepDefOptions = {}, handler = () => {}) {}
-
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
 function fakeHandler(param) {}
 
+function fakeSetWorld(world: WorldLike) {}
+/* eslint-enable */
```
adamlacoste: This is supposed to re-enable any previously disabled rules.  I think it's safe to use here, but the documentation had a big warning label to be careful with this.  If we wanted to be extra cautious we could explicitly re-enable only those rules which were explicitly disabled above.

modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
```diff
@@ -31,6 +46,25 @@ describe("CucumberManager", () => {
...
 
     describe("methods", () => {
+        describe("defineParamType()", () => {
+            let cm;
+
+            beforeEach(() => {
+                cm = new CucumberManager(cucumber, cmOpts);
+
+                vi.spyOn(cm.cucumber, "defineParameterType");
+            });
```
adamlacoste, line 56: This doesn't strictly need to be in a `beforeEach` because we only have one test in this group, but if we were to move it into the test logic and then later add more tests, we'd have to extract it again.

This `beforeEach` logic was copied from the `defineStep()` group below and is repeated again for `defineWorld()` so we could alternatively just move it up a level to the `methods` group and reduce three redundant blocks to one.  I would have done this, but I had a sneaking suspicion that the pre-flight setup needs of these three methods wouldn't necessarily be universal for all future methods we might ever add.  Maybe I should just make it as efficient as it can be for now and let future us take care of the future, but that was the concern that stopped me from consolidating.

modules/@specify-bdd/specify/src/lib/CucumberManager.ts
```diff
@@ -1,12 +1,6 @@
 import assert from "node:assert";
 
-import type { Given, When, Then } from "@cucumber/cucumber";
-
-export interface CucumberLike {
-    Given: typeof Given;
-    When: typeof When;
-    Then: typeof Then;
-}
+import * as Cucumber from "@cucumber/cucumber";
```
adamlacoste, line 3: I realized as I was creating a type based on Cucumber's `World` class that I could do the same with Cucumber itself, and thereby reduce the amount of redundant definition I was doing in here with the `CucumberLike` interface.  You'll see `CucumberLike` down below as a type now, based simply on the amalgamated exports of the Cucumber package.

modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
wild-love: I think it would be a good idea to match the level of explicity to whatever was used to disable the rules in the first place. In this case, itd be listing the rules.

modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
wild-love: I think its fine either way. As you say, future us will handle it 😂

modules/@specify-bdd/specify/src/index.ts
```diff
@@ -25,4 +39,15 @@ export function defineStep(
     cm.defineStep(pattern, handler, options);
 }
 
-export default { defineStep };
+/**
+ * Register a custom world constructor with Specify.  It is STRONGLY recommended
```
wild-love: The warning given here might also be well-placed as a `@remarks`, but Im fine with it either way.

wild-love (changes requested):

modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
adamlacoste: 👍 

- [x] Get specific in ESLint enable comment

modules/@specify-bdd/specify/src/index.ts
adamlacoste: - [x] Move `defineWorld` recommendation into `@remarks` block

wild-love (approved):

PR #246, opened 2026-01-20
addHook method(s)

### LINKED ISSUE #244: Add defineHook method

adamlacoste:
Resolves #244.

### DIFF
--- a/modules/@specify-bdd/specify/src/index.ts
+++ b/modules/@specify-bdd/specify/src/index.ts
@@ -4,8 +4,10 @@ import { config                } from "@/config/all";
 import { CucumberManager, Hook } from "@/lib/CucumberManager";
 
 import type {
+    HookHandler,
     HookOptions,
     ParamTypeOptions,
+    StepDefHandler,
     StepDefOptions,
     StepDefPattern,
     WorldLike,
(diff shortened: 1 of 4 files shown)


modules/@specify-bdd/specify/src/lib/tests/integration/index.test.ts
```diff
@@ -1,15 +1,19 @@
...
+const { mockAfter, mockDefineParameterType, mockGiven, mockSetWorldConstructor } = vi.hoisted(
+    () => {
+        return {
+            "mockAfter":               vi.fn(),
+            "mockDefineParameterType": vi.fn(),
+            "mockGiven":               vi.fn(),
+            "mockSetWorldConstructor": vi.fn(),
+        };
+    },
+);
```
adamlacoste: IMO this formatting is kinda silly -- why not break these objects across multiple lines? -- but Prettier wants it this way and Prettier gets what it wants.

modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
```diff
@@ -46,6 +53,38 @@ describe("CucumberManager", () => {
...
+
+                cm.defineHook(Hook.After, fakeHandler, opts)
+                    .defineHook(Hook.AfterAll, fakeHandler, opts)
+                    .defineHook(Hook.Before, fakeHandler, opts)
+                    .defineHook(Hook.BeforeAll, fakeHandler, opts);
+
+                expect(cm.cucumber.After).toHaveBeenCalledTimes(1);
+                expect(cm.cucumber.AfterAll).toHaveBeenCalledTimes(1);
+                expect(cm.cucumber.Before).toHaveBeenCalledTimes(1);
+                expect(cm.cucumber.BeforeAll).toHaveBeenCalledTimes(1);
```
adamlacoste: I did something a little different here and tested all four permutations of the hook registration inputs.  My thinking was that if we had a non-obvious bug in the switch structure (e.g. the input `Hook.AfterAll` calls `Cucumber.After` or something), we wouldn't necessarily catch it by testing just one path through.

After doing this, I'm inclined to go back to the `defineStep` test and expand it similarly to cover `Given`, `When`, and `Then` testing (rather than what it's testing now, which is just `When`).  However, this would be a bit of scope creep for this ticket, so I thought I'd seek a second opinion first.  IMO it's an easy change that doesn't warrant a separate ticket.

modules/@specify-bdd/specify/src/lib/CucumberManager.ts
```diff
@@ -2,11 +2,24 @@ import assert from "node:assert";
 
 import * as Cucumber from "@cucumber/cucumber";
 
+export enum Hook {
+    BeforeAll,
+    Before,
+    After,
+    AfterAll,
+}
```
adamlacoste, line 12: I waffled a lot on the capitalization of this enum and its properties before settling where I did.  Arguably it should use lower camel case (e.g. `hooks.beforeAll`) because it's not a class, but it looked weird to me.  The official TS documentation on enums uses PascalCase like above, so I defaulted to that until we could discuss.

modules/@specify-bdd/specify/src/lib/CucumberManager.ts
```diff
@@ -56,6 +69,38 @@ export class CucumberManager {
...
+                this.cucumber.AfterAll(options, handler);
+                break;
+            case Hook.Before:
+                this.cucumber.Before(options, handler);
+                break;
+            case Hook.BeforeAll:
+                this.cucumber.BeforeAll(options, handler);
+                break;
+            default:
+                assert.fail(`Invalid hook stage: ${stage}.`);
```
adamlacoste: This should be an unreachable default case, but better safe than sorry.

modules/@specify-bdd/specify/src/index.ts
```diff
@@ -1,19 +1,34 @@
...
 
 import type {
+    HookOptions,
     ParamTypeOptions,
     StepDefOptions,
     StepDefPattern,
     WorldLike,
 } from "@/lib/CucumberManager";
 
+export { Hook } from "@/lib/CucumberManager";
```
adamlacoste: This had to be exported separately to make it available to Specify package consumers.  _Just_ exporting like this without also putting it in the import line above meant it wasn't available for use within this module.  Seems like there ought to be a syntax to do both in the same step. 🤷

adamlacoste:
Moved back to draft because the previous PR (whose branch this work is based on) hasn't been merged yet.

modules/@specify-bdd/specify/src/lib/tests/integration/index.test.ts
```diff
@@ -20,6 +24,14 @@ vi.mock("@cucumber/cucumber", () => {
 
 describe("entrypoint module (integration)", () => {
     describe("exports", () => {
+        describe("defineHook()", () => {
+            it("registers a hook script with Cucumber", () => {
+                defineHook(Hook.After, () => true, {});
```
wild-love: I think this would work, but I'm realizing its not my favorite thing ever 😓

What if we refactor `defineHook` as a static object, and bake in the hook types as methods? Something like:
`DefineHook.after()`?

wild-love: Looking further below, I see the hook methods are using the cucumber instance, so this might not work. But Ill leave it here for discussion.

wild-love (approved):

modules/@specify-bdd/specify/src/lib/tests/integration/index.test.ts
adamlacoste: I don't love the sub-object solution, but I would be OK with adding some convenience methods:

* `defineAfterHook`
* `defineAfterAllHook`
* `defineBeforeHook`
* `defineBeforeAllHook`

Under the hood, we could just have all of these call `this.defineHook` and provide a hard-coded first param.

We could alternatively style these like `defineHookAfterAll` to ensure they alphabetize together, but IMO that's much more cumbersome linguistically, and I think we'd be privileging our own convenience over our users there.

wild-love: I think the convenience methods are a good middle ground, Im in favor. As for the naming, `define...Hook` makes the most sense to me, but I can live with it either way.

modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
wild-love: Hmm... Instead of creating a new ticket, could we necro the "defineStep" branch and make the change there as a separate PR? Seems like the easiest way to make that change without adding it to an unrelated branch.

I understand if you dont want to be hassled though, its not gonna hurt anything to be added to this ticket.

modules/@specify-bdd/specify/src/lib/CucumberManager.ts
wild-love: I like all enums as PascalCase myself (second only to all caps `Hook.BEFOREALL`), so I have no problem with this whatsoever 😂

modules/@specify-bdd/specify/src/lib/tests/integration/index.test.ts
adamlacoste: - [x] Create convenience methods named like `defineAfterHook` which utilize `defineHook`

modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
adamlacoste: - [x] Necro #227 to add more unit testing for different Cucumber methods

modules/@specify-bdd/specify/src/lib/CucumberManager.ts
adamlacoste: I'm also not averse to all caps, though I'd favor shouting snake case (`Hook.BEFORE_ALL`).  But since you're fine with it how it is, I'll leave it and we can revisit this if we decide on a convention for these kinds of things later.

modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
adamlacoste: Opened #251 for this.

modules/@specify-bdd/specify/src/lib/tests/integration/index.test.ts
adamlacoste: After some further discussion, we changed the name of the base method to `addHook` and I've added 6 convenience methods (named variations on `addXHook`) for every supported hook stage, including 2 (`BeforeStep` and `AfterStep`) that I had previously overlooked.

wild-love (approved):

PR #250, opened 2026-01-21
Handler sig bug

### LINKED ISSUE #249: Handler function signatures don't match
Our handler function params in `defineHook` and `defineStepDef` have signatures that don't match the data we're going to need to pass in.  More formalized types that cover the spectrum of valid inputs are needed.

adamlacoste:
Resolves #249.

### DIFF
--- a/modules/@specify-bdd/specify/src/lib/CucumberManager.ts
+++ b/modules/@specify-bdd/specify/src/lib/CucumberManager.ts
@@ -14,6 +14,10 @@ interface ExpressionVariant {
     pattern: RegExp | string;
 }
 
+export interface HookHandler {
+    (...args: any[]): Promise<void> | void; // eslint-disable-line @typescript-eslint/no-explicit-any
+}
+
 export interface HookOptions {
     name?: string;
     tags?: string;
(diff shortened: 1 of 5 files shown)


modules/@specify-bdd/specify/src/lib/CucumberManager.ts
```diff
@@ -14,6 +14,10 @@ interface ExpressionVariant {
     pattern: RegExp | string;
 }
 
+export interface HookHandler {
+    (...args: any[]): Promise<void> | void; // eslint-disable-line @typescript-eslint/no-explicit-any
+}
```
adamlacoste, line 21: It might be possible to get more specific with the params of this function interface.  The function accepts 0-2 params of semi predictable type.  But the specific signature varies depending on what KIND of hook is being registered, and we don't currently distinguish by hook stage.  This might change, so it would be worth revisiting later to give our TS users more specific type hinting.

modules/@specify-bdd/specify/src/lib/CucumberManager.ts
```diff
@@ -32,6 +36,10 @@ export interface ParamTypeOptions {
     useForSnippets?: boolean;
 }
 
+export interface StepDefHandler {
+    (...args: any[]): Promise<void> | void; // eslint-disable-line @typescript-eslint/no-explicit-any
+}
```
adamlacoste, line 43: This one, on the other hand, could be pretty much anything.  The number of input params varies with the format of the pattern string, and while the values extracted from gherkin are initially strings, they are transformed by param definitions into whatever type a user might want.  I think we're stuck with `any` here for the long haul.

wild-love (approved):

adamlacoste:
Conflicts resolved, ready for review.

adamlacoste:
It seems I fixed this in two places and the other one got merged first, so there is nothing left for this PR to do.

PR #251, opened 2026-01-22
More defineStep tests

### LINKED ISSUE #227: Add defineStep method

adamlacoste:
[#227] Added a lot more assertions to defineStep tests validating that Given and Then behavior works just like When does

### DIFF
--- a/modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
+++ b/modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
@@ -71,16 +71,35 @@ describe("CucumberManager", () => {
             beforeEach(() => {
                 cm = new CucumberManager(cucumber, cmOpts);
 
+                vi.spyOn(cm.cucumber, "Given");
                 vi.spyOn(cm.cucumber, "When");
+                vi.spyOn(cm.cucumber, "Then");
             });
 
             describe("registers step def patterns with Cucumber using...", () => {
                 it("a single basic string expression", () => {
+                    cm.defineStep("Given that I have done something with {param}", fakeHandler);
                     cm.defineStep("When I do something with {param}", fakeHandler);
+                    cm.defineStep("Then something should have been done with {param}", fakeHandler);
 
+                    expect(cm.cucumber.Given).toHaveBeenCalledTimes(1);
                     expect(cm.cucumber.When).toHaveBeenCalledTimes(1);
+                    expect(cm.cucumber.Then).toHaveBeenCalledTimes(1);
+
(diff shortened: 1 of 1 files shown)


modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
```diff
@@ -71,83 +71,217 @@ describe("CucumberManager", () => {
...
+                            "Then [I should have done/the user should have done] something with {param}",
+                            fakeHandler,
+                        );
+
+                        expect(cm.cucumber.Given).toHaveBeenCalledTimes(2);
                         expect(cm.cucumber.When).toHaveBeenCalledTimes(2);
+                        expect(cm.cucumber.Then).toHaveBeenCalledTimes(2);
+
+                        expect(cm.cucumber.Given).toHaveBeenCalledWith(
+                            "that (I )did something with {param}", // TODO: this phrasing is weird
```
wild-love, line 222: Whats weird about it? 🤔

wild-love (approved):

modules/@specify-bdd/specify/src/lib/tests/CucumberManager.test.ts
adamlacoste: When we previously discussed the implicit subject feature, I was thinking entirely about "when" steps.  For instance,
```
When the user goes to the store
And the user buys some milk
```
is awkward to read, whereas
```
When the user goes to the store
And buys some milk
```
is much more natural.  HOWEVER, when you consider the implicit subject feature in terms of "Given" steps that use the word "that", it gets weird.
```
Given that the user has gone to the store
And that has bought some milk
```
If "that" were part of the optional clause (`And has bought some milk`), it would be much less awkward/broken sounding.  So I might need to go back and tweak this feature some more.

adamlacoste: Another possibility we could consider is always making our "thats" optional.  I think you usually prefer to phrase steps without them anyway, and you're surely not alone.  If we made them universally optional, the problem goes away. 🤔

wild-love: Always optional sounds like the perfect solution to me :100:

adamlacoste: Created #258 to address this work so I don't inflate the scope of this ticket.

PR #253, opened 2026-01-26
Shell KILLA!!!

wild-love:


features/plugin-cli/steps/shell.feature
```diff
@@ -14,10 +14,22 @@ Feature: Shell Step Definitions
...
             # Then there should be 2 active CLI shells
 
-        # Scenario: Create and destroy a single shell
-        #     Given a CLI shell
-        #     When the user kills a CLI shell
-        #     Then there should be 0 active CLI shells
+        Scenario: Destroy the active shell
+            Given a CLI shell
+            When the user kills the CLI shell
+            # Then there should be 0 active CLI shells
```
wild-love: To implement, or not to implement; that is the question.

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
```diff
@@ -173,6 +179,48 @@ async function execCommandSync(command: string): Promise<void> {
...
+}
+
+/**
+ * Kill a CLI shell by its name.
+ *
+ * @param name - The name of the CLI shell to kill
+ */
+async function killCLIShellByName(name: string): Promise<void> {
+    killCLIShellBySelector.call(this, name);
+}
```
wild-love, line 207: Same strategy here for function overloading as what we used for shell switching.

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
```diff
@@ -185,7 +233,7 @@ async function sendKillSignal(signal: string): Promise<void> {
 
 /**
  * Start a user-specified shell.
- *
```
wild-love, line 188: The linter caught a bunch of these little formatting errors. Interesting that the PR checks dont catch these 🤔

modules/@specify-bdd/plugin-cli/src/lib/SessionManager.ts
```diff
@@ -207,6 +207,31 @@ export class SessionManager {
...
+     * Find the SessionMeta that matches the given selector.
+     *
+     * @param selector - The index or name of the session to find
+     *
+     * @returns The matching SessionMeta
+     *
+     * @throws Error
+     * If there is no matching session
+     **/
+    findSession(selector: number | string): SessionMeta {
```
wild-love, line 220: Basically does exactly what `switchToSession` was doing previously

modules/@specify-bdd/plugin-cli/src/lib/SessionManager.ts
```diff
@@ -330,32 +355,19 @@ export class SessionManager {
...
-        
-        if (!session) {
-            const selectorType = typeof selector === "string" ? "name" : "index";
-            
-            throw new Error(`No session found with ${selectorType}: ${selector}`);
-        }
-
-        this.#activeSession = session;
+    switchToSession(opts: SessionManagerOptions = {}): void {
+        this.#activeSession = opts.sessionMeta;
```
wild-love, line 370: Not a whole lot going on here now, but we could maybe add some checks against invalid arguments if we think there should be some defensive programming here.

features/plugin-cli/steps/shell.feature
wild-love: - [x] Implement

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
adamlacoste: Yeah the differences between `lint` (what the PR checks use) and `lint:fix` have bitten us a few times recently, we should look into what it would take to bring those into alignment.

modules/@specify-bdd/plugin-cli/src/lib/SessionManager.ts
adamlacoste: The only thing I can think to add is a check confirming that `opts.sessionMeta` is a session we are already managing, e.g. by confirming that it exists within `this.sessions`.  It's highly unlikely that anybody's going to ask us to switch to a session we aren't managing (where would they get it?), but if we want to be SUPER careful, it could prevent some potentially odd behavior.

adamlacoste: Oh, actually there is one other thing.  What happens if the caller doesn't provide an `opts` param?  `opts.sessionMeta` would be undefined and therefore so would `this.#activeSession`.  That's probably not the behavior we want, so it would behoove us to verify that they've actually passed us a session to switch to.

adamlacoste (commented):
Had a few thoughts, but for the most part this is looking good.  Will re-review after changes are in.

modules/@specify-bdd/plugin-cli/src/lib/SessionManager.ts
wild-love: - [x] Add value check for the options object

adamlacoste (approved):

PR #257, opened 2026-01-28
Move SpecifyWorld to lib

### LINKED ISSUE #255: Move SpecifyWorld class into lib module
The `SpecifyWorld` class which extends Cucumber's `World` currently lives in the cucumber/world.ts module.  We need to move it to lib, where it belongs, so that it can be unit tested and exported by the index entrypoint module.

adamlacoste:
Resolves #255.


modules/@specify-bdd/specify/src/lib/tests/SpecifyWorld.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    describe("constructor", () => {
+        it("passes options through to the Cucumber World constructor", () => {
+            expect(World).toHaveBeenCalledTimes(1);
+            expect(World).toHaveBeenCalledWith(opts);
+        });
+
+        it("initializes a new QuickRef instance", () => {
+            expect(QuickRef).toHaveBeenCalledTimes(1);
+            expect(QuickRef).toHaveBeenCalledWith();
+        });
```
adamlacoste, line 40: As we discussed on the call, this test is somewhat redundant with the `quickRef` property testing below.  However, I feel like verifying that the class instance has the expected state after construction is valid in its own right.  I'm not dead-set on keeping this method, but my preference is to keep it.

modules/@specify-bdd/specify/src/index.ts
```diff
@@ -13,6 +13,8 @@ import type {
     WorldLike,
 } from "@/lib/CucumberManager";
 
+export { SpecifyWorld } from "@/lib/SpecifyWorld";
```
adamlacoste, line 16: Is this something we should add an integration test for?  It is part of the public API for Specify, but we are literally just passing through a class that is already unit tested on its own.

wild-love: This is pretty neat syntax! I would lean towards not testing it, since its just a passthrough and it contains no logic.

wild-love (approved):

PR #260, opened 2026-02-02
GH Actions version bumps

adamlacoste:
Updated github action workflow versions that had fallen out of date

### DIFF
--- a/.github/workflows/release-stage-1-test.yml
+++ b/.github/workflows/release-stage-1-test.yml
@@ -11,6 +11,8 @@ jobs:
     name: Test active Node.js LTS version
     uses: ./.github/workflows/release-stage-1a-test-node-version.yml
     with:
+      audit: true
+      lint: true
       node_version: ${{ vars.NODE_LTS_ACTIVE_VERSION }}
       store_artifacts: true
   test-maintenance:
--- a/.github/workflows/release-stage-1a-test-node-version.yml
+++ b/.github/workflows/release-stage-1a-test-node-version.yml
@@ -55,7 +55,7 @@ jobs:
         run: pnpm test
       - name: Upload build artifacts
         if: ${{ inputs.store_artifacts == true }}
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v6
         with:
           name: artifact_build
           path: ./modules/@specify-bdd/*/dist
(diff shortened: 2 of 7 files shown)


.github/workflows/release-stage-1-test.yml
```diff
@@ -11,6 +11,8 @@ jobs:
     name: Test active Node.js LTS version
     uses: ./.github/workflows/release-stage-1a-test-node-version.yml
     with:
+      audit: true
+      lint: true
```
adamlacoste, line 15: These params tell the 1a substage script to audit and lint JUST the current LTS version, not the older LTS versions.  Doing it three times would be redundant since it's the same code and dependency list either way.

.github/workflows/release-stage-1a-test-node-version.yml
```diff
@@ -20,19 +28,25 @@ jobs:
...
         with:
           cache: pnpm
           node-version: ${{ inputs.node_version }}
           registry-url: "https://registry.npmjs.org"
+      - name: Lint
+        if: ${{ inputs.lint == true }}
+        run: pnpm lint
+      - name: Audit
+        if: ${{ inputs.audit == true }}
+        run: pnpm audit --prod --audit-level=moderate
```
adamlacoste, line 49: These steps were previously part of stage 2 (tag), probably because I didn't want to parallelize them, but that's better solved with the conditional logic here.  Semantically, these steps are testing whether the code in the source branch (`develop`, usually) is good to release so they rightly belong in stage 1.

.github/workflows/release-stage-2-tag.yml
```diff
@@ -49,19 +49,16 @@ jobs:
...
       - name: Download build artifacts
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@v7
         with:
           name: artifact_build
-      - name: Lint
-        run: pnpm lint
-      - name: Audit
-        run: pnpm audit --prod --audit-level=moderate
+          path: modules/@specify-bdd/
```
adamlacoste, line 61: This was the real fix needed.  The artifact upload happening in stage 1 was working with a project root-relative path to the dist dirs, so I mistakenly assumed that the internal zip file paths of all dist files would be archived relative to the project root.  Not so, apparently!  Instead the "least common ancestor" is used as the internal archive path root.  This extra bit of specification makes sure that the artifacts are imported back to their correct places in the repo dir structure.

wild-love (approved):

PR #263, opened 2026-02-03
Port Specify core to CucumberManager

### LINKED ISSUE #247: Port Specify core package to CucumberManager

adamlacoste:
Resolves #247.


modules/@specify-bdd/specify/src/cucumber/step_definitions/demo.ts
```diff
@@ -3,17 +3,20 @@
...
+import assert from "node:assert/strict";
 
-When("this step fails", fail);
-When("this step passes", pass);
+import { defineStep } from "@/index";
 
-Then("this step should fail", fail);
-Then("this step should pass", pass);
+defineStep(
+    ["Given that this step passes", "When this step passes", "Then this step should pass"],
```
adamlacoste, line 11: Not how I'd prefer to format this, but we are thralls to the prettier rules. 😒

modules/@specify-bdd/specify/src/cucumber/step_definitions/testing.ts
```diff
@@ -35,7 +41,7 @@ async function passAfterDelay(delay: number): Promise<void> {
  *
  * @param workers - The expected number of parallel workers
  */
-async function passIfNWorkers(workers: number): Promise<void> {
+function passIfNWorkers(workers: number): void {
```
adamlacoste, line 43: I'm not sure if we have discussed this before; maybe we decided at some point to make it a convention that all step def handler functions should be async by default?  But looking at it with fresh eyes, I thought it was quite silly given what we're doing with them, and since our codebase is meant to be exemplary for Specify users, I thought it best to do the sensible thing that needs no explanation when read by third parties.

modules/@specify-bdd/specify/src/cucumber/support/hooks.ts
```diff
@@ -4,67 +4,77 @@
...
-import { error                    } from "node:console";
+import { globby        } from "globby";
+import path              from "node:path";
+import { pathToFileURL } from "node:url";
+import { error         } from "node:console";
 
 import { config          } from "@/config/all";
 import { CucumberManager } from "@/lib/CucumberManager";
 
+import { addAfterScenarioHook, addBeforeAllHook, addBeforeScenarioHook } from "@/index";
```
adamlacoste: I think this `import` was set off on its own because when I wrote it, the methods were broken across three lines.  Prettier "fixed" that for me and now this is breaking our convention of keeping single-line internal imports together.  I could delete the blank line above, but doing so mean I then need to whitespace the shit out of the two imports above that to make the braces line up.

TLDR I hate prettier.

modules/@specify-bdd/specify/src/cucumber/support/hooks.ts
```diff
@@ -4,67 +4,77 @@
...
+
 import type { JsonObject } from "type-fest";
 
 const cwd = process.cwd(),
     pickleJar = {}; // keep a store of pickles in the module scope so we can track test case attempts
 
-let refsMods = [];
+let refsModules = [];
```
adamlacoste, line 26: When I returned to this hook code, I initially struggled to understand what the variable `refsMods` was being used for until I stopped to read the method thoroughly.  That seemed like a red flag that the abbreviation I used was obfuscating its purpose, so I thought it would be best to clarify it.

modules/@specify-bdd/specify/src/cucumber/support/hooks.ts
```diff
@@ -4,67 +4,77 @@
...
-});
+        // warn if CucumberManager registered any duplicate step patterns
+        // but only do it once per run (even if running multiple parallel workers)
+        if (!process.env.CUCUMBER_WORKER_ID || process.env.CUCUMBER_WORKER_ID === "0") {
+            Object.entries(CucumberManager.getInstance().patterns)
+                .filter((entry) => entry[1] > 1)
+                .forEach((entry) =>
+                    error("WARNING: pattern '%s' was registered %d times.", entry[0], entry[1]),
+                );
+        }
```
adamlacoste, line 51: All of the logic within the `BeforeAll` hook callback function is unchanged, apart from the renaming of `refsModules`.  It's just prettier's whitespace changes that make it part of the diff.

modules/@specify-bdd/specify/src/cucumber/support/hooks.ts
```diff
@@ -4,67 +4,77 @@
...
+    },
+    { "name": "Core before all hook" },
```
adamlacoste, line 53: I think putting an optional `options` object at the end of a function's param list is the correct and appropriate thing to do, but I'll be the first to admit that in this case, it does lead to the unfortunate side effect of putting the hook's name all the way down at the bottom of the block, burying the lede somewhat.  We could mitigate this by declaring a variable for the options object above the hook block, but it's not as svelte as the original configuration, I'll admit.

modules/@specify-bdd/specify/src/cucumber/support/hooks.ts
adamlacoste: The same is true for the Before and After hooks below.

test/gherkin/parallel/workers1.feature
adamlacoste, line 1: These changes (and the ones in the workers2.feature file after this) are an attempt to correct what I saw as very awkward phrasing in one of the step definitions above.  "When" steps describe user actions or inputs, whereas "Then" steps describe system reactions or outputs.  "When this step passes if there is 1 parallel worker" is clearly testing an output, but using a "When" step to do it.  I suspect that this was done because a scenario is usually expected to describe a user action followed by a system reaction, so just starting with a "Then" step seemed weird.  I've chosen to address the weirdness with implicit (commented out) user action "When" steps, wherein we explicitly state that the user input is running the command a certain way.

If this seems like a bad change, I'm open to being talked out of it.

modules/@specify-bdd/specify/src/cucumber/step_definitions/demo.ts
```diff
@@ -3,17 +3,20 @@
...
-import { Given, When, Then } from "@cucumber/cucumber";
-import assert                from "node:assert/strict";
 
-Given("that this step fails", fail);
-Given("that this step passes", pass);
+import assert from "node:assert/strict";
 
-When("this step fails", fail);
-When("this step passes", pass);
+import { defineStep } from "@/index";
```
wild-love: I can't say Im partial to this `@/index`, its about as nondescript as it gets 👀 

Is there something else we can call this? Maybe alias it to `@/specify`?

modules/@specify-bdd/specify/src/cucumber/support/hooks.ts
wild-love: Yeah I agree that it is a bit cumbersome this way, but I dont think its bad enough to demand a refactor 🤷

modules/@specify-bdd/specify/src/cucumber/support/world.ts
```diff
@@ -1,5 +1,4 @@
-import { setWorldConstructor } from "@cucumber/cucumber";
-
+import { setWorld     } from "@/index";
```
wild-love: Good golly I do not care for this import module name 😵

test/gherkin/parallel/workers1.feature
wild-love: I find both options a bit awkward, but the new way's `then` step makes the most sense to me, and we already use comments for explanations in the rest of the codebase. I suppose I can learn to love it, in time 😂

modules/@specify-bdd/specify/src/cucumber/step_definitions/demo.ts
adamlacoste: I don't know how these TS aliases work.  Maybe it's possible?  Since `@/index` refers to the `src/index.ts` module, it seems pretty clear to me. 🤷

modules/@specify-bdd/specify/src/cucumber/support/hooks.ts
adamlacoste: - [x] group index import in hooks module with other internal imports 😒

modules/@specify-bdd/specify/src/cucumber/step_definitions/demo.ts
adamlacoste: Per our discussion on a call, we're going to try using the whole package name import (`@specify-bdd/specify`) from within the package itself; this should clear up any ambiguity, assuming it works.

- [x] Replace all `@/index` imports with the full package name

adamlacoste:
All feedback resolved, ready for re-review.

wild-love (approved):

Issue #258, opened 2026-01-29
Update behavior specs to utilize enhanced syntax


adamlacoste:
We should update our Gherkin behavior specs and step definitions to utilize new features added in the enhanced expressions release.  In particular:

- [x] Support both first and third person subjects
- [x] Use implicit subjects where applicable
- [x] Make "that" optional in all "Given" steps
- [x] If the above makes "Given"/"When" patterns ambiguous, convert "Given" to use past tense

PR #264, opened 2026-02-04
Port CLI to CM

### LINKED ISSUE #248: Port CLI plugin package to CucumberManager

adamlacoste:
Resolves #248.

### DIFF
--- a/modules/@specify-bdd/plugin-cli/package.json
+++ b/modules/@specify-bdd/plugin-cli/package.json
@@ -21,6 +21,7 @@
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "dependencies": {
+    "@specify-bdd/specify": "workspace:*",
     "ps-list": "^9.0.0",
     "tree-kill": "^1.2.2",
     "validate-npm-package-name": "^6.0.2"
@@ -29,9 +30,6 @@
     "@types/validate-npm-package-name": "^4.0.2",
     "tsc-alias": "^1.8.16",
     "typescript": "^5.9.3"
-  },
-  "peerDependencies": {
-    "@cucumber/cucumber": "^11.3.0"
   },
   "repository": {
     "type": "git",
--- a/modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/file_system.ts
+++ b/modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/file_system.ts
@@ -50,10 +50,7 @@ defineStep(
 );
 
 defineStep(
-    [
-        "Then the {filePath} file path should exist",
-        "Then the {ref} file path should exist",
-    ],
+    ["Then the {filePath} file path should exist", "Then the {ref} file path should exist"],
     verifyFilePathExists,
 );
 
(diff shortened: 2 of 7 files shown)


modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/file_system.ts
```diff
@@ -4,34 +4,55 @@
...
-When("a/the user deletes the {filePath} file", deleteFile);
-
-Then("the {filePath} file content should be empty", verifyFileIsEmpty);
-Then("the {ref} file content should be empty", verifyFileIsEmpty);
-
-Then("the {filePath} file content should match {ref}", verifyFileContent);
-Then("the {ref} file content should match {ref}", verifyFileContent);
-
-Then("the {filePath} file path should exist", verifyFilePathExists);
-Then("the {ref} file path should exist", verifyFilePathExists);
```
adamlacoste, line 34: For the purposes of keeping scope under control for this ticket, I deliberately avoided doing a lot of "cleanup" on our existing patterns as part of this work.  All I did here was swap Cucumber's `Given`, `When`, and `Then` methods for our `defineStep` and consolidated patterns wherever multiple calls were using the same handler function.  I'll be back to tidy up the patterns and make them more efficient with #258.

modules/@specify-bdd/plugin-cli/src/cucumber/support/hooks.ts
```diff
@@ -4,37 +4,42 @@
  * Custom Cucumber hooks enabling CLI testing.
  */
 
-import { After, AfterAll, Before, BeforeAll } from "@cucumber/cucumber";
-import * as fs                                from "node:fs/promises";
+import { addAfterAllHook, addAfterScenarioHook, addBeforeScenarioHook } from "@specify-bdd/specify";
```
adamlacoste, line 7: For some stupid reason prettier wants this whole long ass thing on one line.

modules/@specify-bdd/plugin-cli/src/cucumber/support/hooks.ts
```diff
@@ -4,37 +4,42 @@
...
+        if (this.parameters.cliCleanup) {
+            for (const file of cliFiles) {
+                promises.push(fs.unlink(file));
+            }
         }
-    }
 
-    return Promise.all(promises);
-});
+        await Promise.all(promises);
```
adamlacoste, line 42: We were returning this promise for reasons unclear, but nothing is or can be done with it.  (It was probably my reflex of making a function return something rather than nothing.)  But our `addXHook` methods expect the callback to return void, so we just await the promise now instead.

modules/@specify-bdd/plugin-cli/package.json
```diff
@@ -21,6 +21,7 @@
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "dependencies": {
+    "@specify-bdd/specify": "workspace:*",
```
adamlacoste, line 24: We no longer depend on `@cucumber/cucumber` here, but we do need to be able to import methods (e.g. `defineStep`, `addXHook`) from specify core's entrypoint.

wild-love (approved):

PR #268, opened 2026-02-10
Cleanup Specify step defs

### LINKED ISSUE #265: Cleanup Specify step defs

adamlacoste:
Resolves #265.

### DIFF
--- a/modules/@specify-bdd/specify/src/cucumber/step_definitions/demo.ts
+++ b/modules/@specify-bdd/specify/src/cucumber/step_definitions/demo.ts
@@ -8,12 +8,12 @@ import { defineStep } from "@specify-bdd/specify";
 import assert         from "node:assert/strict";
 
 defineStep(
-    ["Given that this step passes", "When this step passes", "Then this step should pass"],
+    ["Given (that )this step passes", "When this step passes", "Then this step should pass"],
     pass,
 );
 
 defineStep(
-    ["Given that this step fails", "When this step fails", "Then this step should fail"],
+    ["Given (that )this step fails", "When this step fails", "Then this step should fail"],
     fail,
 );
 
--- a/modules/@specify-bdd/specify/src/cucumber/step_definitions/demo.ts
+++ b/modules/@specify-bdd/specify/src/cucumber/step_definitions/demo.ts
@@ -8,12 +8,12 @@ import { defineStep } from "@specify-bdd/specify";
 import assert         from "node:assert/strict";
 
 defineStep(
-    ["Given (that )this step passes", "When this step passes", "Then this step should pass"],
+    ["Given (that )this step has passed", "When this step passes", "Then this step should pass"],
     pass,
 );
 
 defineStep(
-    ["Given (that )this step fails", "When this step fails", "Then this step should fail"],
+    ["Given (that )this step has failed", "When this step fails", "Then this step should fail"],
     fail,
 );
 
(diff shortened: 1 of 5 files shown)


modules/@specify-bdd/specify/src/cucumber/step_definitions/testing.ts
```diff
@@ -7,28 +7,26 @@
 import { defineStep } from "@specify-bdd/specify";
 import assert         from "node:assert/strict";
 
-defineStep("Given that this step passes after {float} seconds", passAfterDelay);
-defineStep("When the user waits for {float} second(s)", waitForTime, { "timeout": 60000 });
```
adamlacoste, line 11: These two step defs were using two different handler functions, `passAfterDelay` and `waitForTime`, that were functionally doing the exact same thing.  I decided to make `waitForTime` the canonical handler and combined both patterns to use that one.

modules/@specify-bdd/specify/src/cucumber/step_definitions/testing.ts
adamlacoste, line 70: Here is `waitForTime`, for reference.

modules/@specify-bdd/specify/src/cucumber/step_definitions/testing.ts
```diff
@@ -7,28 +7,26 @@
...
 );
 
-/**
- * Always passes after delay.
- *
- * @param delay - The number of seconds to wait before passing
- */
-async function passAfterDelay(delay: number): Promise<void> {
-    return new Promise((resolve) => setTimeout(resolve, delay * 1000));
-}
```
adamlacoste, line 30: Although it wasn't changed, I've staked out the `waitForTime` method with a comment below so they can be easily compared.

modules/@specify-bdd/specify/src/cucumber/step_definitions/demo.ts
```diff
@@ -8,12 +8,12 @@ import { defineStep } from "@specify-bdd/specify";
 import assert         from "node:assert/strict";
 
 defineStep(
-    ["Given that this step passes", "When this step passes", "Then this step should pass"],
+    ["Given (that )this step has passed", "When this step passes", "Then this step should pass"],
```
adamlacoste, line 11: I had to change the phrasing of this step because when the word "that" is made optional (a convention we recently decided to adopt universally), the "given" and "when" forms of this step become potentially ambiguous.  ("this step passes" matches both.)  In cases like this, I convention I intend to follow is to make the "given" for a past-tense phrase ("given that a thing has already happened") while keeping the "when" step in present-tense.  I think this semantically supports the idea that given steps represent preconditions for the test, rather than actions being taken as part of the test.

I don't think this past-tense thing is something we necessarily have to be super militant about and there may be some cases where given steps read more naturally in present tense.  But in cases where "given" and "when" forms can be ambiguous but for the "that", I think it makes sense to use tense as the standard way of differentiating them.

wild-love: I like the idea of adopting past-tense as the default "given" tense 👍

modules/@specify-bdd/specify/src/cucumber/step_definitions/testing.ts
```diff
@@ -7,28 +7,26 @@
...
 import assert         from "node:assert/strict";
 
-defineStep("Given that this step passes after {float} seconds", passAfterDelay);
-defineStep("When the user waits for {float} second(s)", waitForTime, { "timeout": 60000 });
 defineStep("Then there should be (only ){int} parallel worker(s)", passIfNWorkers);
 
 defineStep(
     [
-        "Given that this step passes on the {ordinal} attempt",
+        "Given (that )this step has passed after {float} seconds",
```
wild-love, line 14: The phrasing for this is a bit odd to me (what are we comparing the "after seconds" to, since its already passed in the past? 🤔 ), but I'm unable to come up with any suggestions on how to fix it 🤷

wild-love (approved):

PR #269, opened 2026-02-11
Cleanup CLI step defs

### LINKED ISSUE #266: Cleanup CLI plugin step defs

adamlacoste:
Resolves #266.

### DIFF
--- a/features/plugin-cli/steps/shell.feature
+++ b/features/plugin-cli/steps/shell.feature
@@ -47,14 +47,14 @@ Feature: Shell Step Definitions
         Scenario: Create alternative shell type with a name
             Given a "bash" CLI shell named "test-shell"
             And another CLI shell
-            When a user switches to the CLI shell named "test-shell"
+            When the user switches to the CLI shell named "test-shell"
             And the user runs the command "echo $0"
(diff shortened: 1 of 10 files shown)


features/plugin-cli/steps/shell.feature
adamlacoste, line 1: In our move to supporting first and third person as a convention, we're dropping support for third person with an indefinite article ("_a_ user").  Small price to pay, IMO.  it's technically possible to support both definite and indefinite articles, it would be much more verbose for probably very little utility.  We might be able to further augment our enhanced syntax to make a less verbose form possible, but I don't know that it's worth the effort right now.

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/file_system.ts
adamlacoste, line 1: Throughout these step defs, the only change consistently needed was to convert third person subjects with alternative definite or indefinite articles (e.g. `a/the user sings a song`) with enhanced alternative first or third person (definite) syntax (e.g. `[I sing/the user sings] a song`).

wild-love (approved):

PR #276, opened 2026-03-03
Configurable work path

### LINKED ISSUE #262: Post-publish smoke test
We are not currently testing that our published packages work as intended, which is a problem because we recently discovered flaws in the automated release process which probably should have been caught.  We need to add a sixth stage which runs the published package through automated tests to ensure that it is fully functional.  If the demo mode is already implemented, this would be a good use for it, but we may also want to consider a more comprehensive regression test.

### LINKED ISSUE #275: Configurable working directory
The current working directory for test runtime should be a configurable value, instead of simply defaulting to the CWD when the Specify command is executed.  As with all other config values, this should be overridable by the user.  The working directory config should affect the _tests_, not Specify operations surrounding the tests.  For instance, if the configured working directory is "./test" but their refs file is located at "./specify.refs.json", the refs file should still be detected and loaded.

adamlacoste:
Resolves #275.

This changeset adds a new config setting, currently named `paths.working` (but we should discuss that further), which changes the default context in which Cucumber runs its tests.  By default, it's the process's current working directory, so anybody without an override config should see everything continue to work the same as it did before.  This change was necessary to implement in support of #262, which needs to run Specify from one directory but borrows the context of another when executing its tests.

### DIFF
--- a/features/plugin-cli/steps/file_system.feature
+++ b/features/plugin-cli/steps/file_system.feature
@@ -11,5 +11,5 @@ Feature: File System Step Definitions
             And a CLI shell
 
         Scenario:
-            When the user changes the working directory to "./test"
-            # Then the working directory will include "assets" 
+            When the user changes the working directory to "./gherkin"
+            # Then the working directory will include "binary" 
--- a/features/test/basic.feature
+++ b/features/test/basic.feature
@@ -7,7 +7,6 @@ Feature: Basic Test Execution
         Given that the "@specify-bdd/specify" NPM package is installed
         And that the "@specify-bdd/plugin-cli" NPM package is installed
         And a CLI shell
-        And that the working directory is "./test"
 
     Rule: The run should succeed only if all tests pass
 
--- a/modules/@specify-bdd/specify/src/config/cucumber.config.ts
+++ b/modules/@specify-bdd/specify/src/config/cucumber.config.ts
@@ -1,3 +1,5 @@
+import path from "node:path";
+
 import type { IConfiguration } from "@cucumber/cucumber/api";
 
 export type CucumberConfig = Partial<IConfiguration>;
@@ -7,6 +9,8 @@ export const cucumber: CucumberConfig = {
     "retryTagFilter":  "@flaky or @retry",
     "tags":            "not @skip",
     "worldParameters": {
+        // The path of the refs file relative to the project root
+        "refsPath": path.resolve("./specify.refs.json"),
         // similar value as to what you'd see if you ran $PATH in your terminal,
         // but there is also some node-specific path info added to it.
         "userPath": process.env.PATH,
--- a/modules/@specify-bdd/specify/src/config/cucumber.config.ts
+++ b/modules/@specify-bdd/specify/src/config/cucumber.config.ts
@@ -10,7 +10,7 @@ export const cucumber: CucumberConfig = {
     "tags":            "not @skip",
     "worldParameters": {
         // The path of the refs file relative to the project root
-        "refsPath": path.resolve("./specify.refs.json"),
+        "refsPath": path.resolve("./*.refs.json"),
         // similar value as to what you'd see if you ran $PATH in your terminal,
         // but there is also some node-specific path info added to it.
         "userPath": process.env.PATH,
--- a/modules/@specify-bdd/specify/src/config/paths.config.ts
+++ b/modules/@specify-bdd/specify/src/config/paths.config.ts
@@ -12,4 +12,5 @@ export const paths: PathsConfig = {
     "logs":    "./logs/specify",
     "refs":    "./specify.refs.json",
     "rerun":   `${os.tmpdir()}/specify/rerun.txt`,
+    "working": ".",
 };
--- a/modules/@specify-bdd/specify/src/config/paths.config.ts
+++ b/modules/@specify-bdd/specify/src/config/paths.config.ts
@@ -4,13 +4,11 @@ export type PathsConfig = {
     [key: string]: string;
     gherkin: string;
     logs: string;
-    refs: string;
 };
 
 export const paths: PathsConfig = {
     "gherkin": "./features/**/*.feature",
     "logs":    "./logs/specify",
-    "refs":    "./specify.refs.json",
     "rerun":   `${os.tmpdir()}/specify/rerun.txt`,
     "working": ".",
 };
--- a/modules/@specify-bdd/specify/src/config/paths.config.ts
+++ b/modules/@specify-bdd/specify/src/config/paths.config.ts
@@ -7,8 +7,8 @@ export type PathsConfig = {
 };
 
 export const paths: PathsConfig = {
-    "gherkin": "./features/**/*.feature",
-    "logs":    "./logs/specify",
-    "rerun":   `${os.tmpdir()}/specify/rerun.txt`,
-    "working": ".",
+    "gherkin":              "./features/**/*.feature",
+    "logs":                 "./logs/specify",
+    "rerun":                `${os.tmpdir()}/specify/rerun.txt`,
+    "testExecutionContext": ".",
 };
--- a/modules/@specify-bdd/specify/src/cucumber/support/hooks.ts
+++ b/modules/@specify-bdd/specify/src/cucumber/support/hooks.ts
@@ -27,7 +27,7 @@ let refsModules = [];
 
 addBeforeAllHook(
     async function () {
-        const modulePaths = await globby(path.join(cwd, "*.refs.json"), {
+        const modulePaths = await globby(this.parameters.refsPath, {
             "absolute":  true,
             "onlyFiles": true,
         });
@@ -67,10 +67,13 @@ addBeforeScenarioHook(
         // initialize an empty attempt object with a unique ID
         this.pickle.attempts[data.testCaseStartedId] = {};
 
+        // reset to the working directory we started with, if necessary
+        if (process.cwd() !== cwd) {
+            process.chdir(cwd);
+        }
+
         // initialize the file system namespace
-        this.fs = {
-            "cwd": process.cwd(),
-        };
+        this.fs = { cwd };
     },
     { "name": "Core before scenario hook" },
 );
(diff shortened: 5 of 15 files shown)


features/plugin-cli/steps/file_system.feature
adamlacoste, line 1: This is the only feature that still needs to change directory, since it is explicitly testing the change dir functionality.

features/test/basic.feature
adamlacoste, line 1: The rest of these were changing directory just as feature-level background step precondition setup, which the new config setting negates the need for.

modules/@specify-bdd/specify/src/config/cucumber.config.ts
adamlacoste, line 1: One problem I ran into while implementing this feature was that after it was working and Cucumber was operating in the context of the configured working directory, it lost the ability to find the refs file in the project root (since that was no longer the WD).  The way I've chosen to fix this is to add `refsPath` as a Cucumber world param, which can be set in the outer Specify context (where the WD is still the project root) and provide an absolute path to Cucumber so the difference in WD isn't a problem.

modules/@specify-bdd/specify/src/config/paths.config.ts
adamlacoste, line 1: The `refs` path defined here wasn't being used for anything, strangely.  In our Cucumber `BeforeAll` hook (next file), we had a hard-coded glob pattern in use instead.  I decided to remove this because it wasted about 45 minutes of my time trying to figure out why it was there, and it doesn't need to waste anyone else's.

modules/@specify-bdd/specify/src/config/paths.config.ts
```diff
@@ -4,12 +4,11 @@ export type PathsConfig = {
...
     logs: string;
-    refs: string;
 };
 
 export const paths: PathsConfig = {
     "gherkin": "./features/**/*.feature",
     "logs":    "./logs/specify",
-    "refs":    "./specify.refs.json",
     "rerun":   `${os.tmpdir()}/specify/rerun.txt`,
+    "working": ".",
```
adamlacoste: Right now this config setting is named `working`, as in the working directory in which we test.  But this is not super intuitive to me anymore, based on how this feature has developed.  It doesn't change the Node process's working directory anymore, it ONLY changes the WD for the Cucumber subprocess.  I think we should rename it to better express that.

@wild-love suggested `context`, which I think is a definite improvement, but I think it could still use more specificity.  `cucumberContext` or `testContext` might be better.  Open to more ideas here.  I think this should change, I'm just not sure to what.

modules/@specify-bdd/specify/src/cucumber/support/hooks.ts
```diff
@@ -27,7 +27,7 @@ let refsModules = [];
 
 addBeforeAllHook(
     async function () {
-        const modulePaths = await globby(path.join(cwd, "*.refs.json"), {
+        const modulePaths = await globby(this.parameters.refsPath, {
```
adamlacoste, line 29: Here the hard-coded glob pattern that assumes refs will be found in the current working directory is replaced with the world parameter `refPath` defined in the cucumber config above.  This fixes the issue of refs not loading if Cucumber's configured working directory is not the project root.

modules/@specify-bdd/specify/src/cucumber/support/hooks.ts
```diff
@@ -68,9 +68,7 @@ addBeforeScenarioHook(
         this.pickle.attempts[data.testCaseStartedId] = {};
 
         // initialize the file system namespace
-        this.fs = {
-            "cwd": process.cwd(),
-        };
+        this.fs = { cwd };
```
adamlacoste, line 70: Not sure why we were re-checking the CWD every scenario, it should not change throughout runtime.  If it did, that would be a form of test pollution that we should prevent/ignore.

modules/@specify-bdd/specify/src/lib/tests/TestCommand.test.ts
```diff
@@ -260,6 +260,19 @@ describe("TestCommand", () => {
...
+                    const debugOpts = { ...userOpts, "debug": true, "workingPath": "./test" };
+                    const cmd       = new TestCommand(debugOpts);
+
+                    mockRunCucumber.mockResolvedValueOnce({ "success": true });
+
+                    const res = await cmd.execute(userArgs);
+
+                    expect(res.debug.args).toBe(userArgs);
+                    expect(res.debug.cucumber.runEnvironment).toEqual({ "cwd": debugOpts.workingPath, "debug": debugOpts.debug });
+                });
```
adamlacoste, line 282: This test validates that config settings like debug mode and working path are being faithfully passed through to Cucumber when we run it.

modules/@specify-bdd/specify/src/lib/Command.ts
adamlacoste, line 1: Originally I wanted the working path to be something that was handled in the base `Command` class because it seemed like something that might have applications in a variety of commands.  However, after I discovered that `Command.execute` does nothing and had to move that logic into `TestCommand`, it made more sense to confine the working directory usage to Cucumber specifically.  So now, I'm not so sure whether it belongs here in `Command` or ought to move to `TestCommand` (the only place it does anything) instead.  Opinions welcome.

modules/@specify-bdd/specify/src/lib/TestCommand.ts
```diff
@@ -133,7 +135,7 @@ export class TestCommand extends Command {
         try {
             const cucumberConfig    = await this.#buildCucumberConfig(userArgs);
             const cucumberRunConfig = await CucumberAPI.loadConfiguration(cucumberConfig);
-            const cucumberEnv       = { "debug": this.debug };
+            const cucumberEnv       = { "cwd": this.workingPath, "debug": this.debug };
```
adamlacoste: This `cucumberEnv` variable gets passed into the `CucumberAPI.runCucumber` method as param # 2, so it literally only affects Cucumber and the rest of Specify's logic operates in the working directory we started from.

specify.config.json
```diff
@@ -5,7 +5,8 @@
     },
     "debug": false,
     "paths": {
-        "logs": "./logs"
+        "logs": "./logs",
+        "working": "./test"
```
adamlacoste: This override ensures that the Specify project runs all of its Cucumber tests in the context of our `./test` sandbox, negating the need for each feature to change directory to avoid config pollution when running Specify-in-Specify.

modules/@specify-bdd/specify/src/config/paths.config.ts
wild-love: I think increasing the specificity is a good idea. My thoughts on your suggestions;
- `cucumberContext` - about as specific as it gets, but its not great if/when we support alternative testing tools
- `testContext` - very generic and tool-agnostic, but it not immediately obvious to me what it means. A comment would help a lot.

Of the two options presented, I like `testContext` + comment. An alternative might be `testEnvContext` or `testExecutionContext`. I think all of those are solidly ok.

modules/@specify-bdd/specify/src/cucumber/support/hooks.ts
wild-love: I think the previous implementation _was_ preventing/ignoring cwd changes 👀 This new version would let those changes persist, if they exist. But it may not matter if we intend to never let it change in the first place.

modules/@specify-bdd/specify/src/lib/Command.ts
wild-love: Im a fan of keeping details at the lowest possible level needed, so keeping it in `TestCommand` is fine by me 👍

modules/@specify-bdd/specify/src/config/paths.config.ts
adamlacoste: I can get behind `testExecutionContext`. 👍 

- [x] Change `working` config setting to `testExecutionContext` in all usage locations

modules/@specify-bdd/specify/src/cucumber/support/hooks.ts
adamlacoste: What the old logic was doing was checking what the current CWD is at the start of each scenario, and using that to initialize the Node.js spawn process (which is silly, because it defaults to the parent process CWD).  So if we're running tests serially and Scenario A changes the working directory to `./foo`, when Scenario B starts, it does so in the context of `./test`.  Whereas if the tests ran in parallel and Scenario B started before Scenario A's working directory change, its operating context would be different and you could potentially get a whole different outcome.  IMO that's a form of test pollution that we should guard against.

With the logic change above, the `cwd` value we capture at module load time remains constant throughout the life of the process and all scenarios that run within this process (which would be all tests if running serially, or the subset delegated to this worker if running in parallel) will share that initial `cwd` value as their starting place.  A `process.changedir()` could still change the process's actual CWD in that period, so the potential for risk is not completely mitigated.  But our shell logic ignores the actual CWD in favor of this cached World value `this.fs.cwd` when initializing isubprocesses, so I think we're guarding against this as well as we can for the time being.

modules/@specify-bdd/specify/src/lib/Command.ts
adamlacoste: Sorry that answer's not clear to me because I think that "lowest" could be interpreted multiple ways, and you refer to "keeping" it in `TestCommand` but that's not where these values currently reside.  Should we...
1. keep the `workingPath` option support in the base `Command` class so it will be available to use with future XCommand subclasses we may add in the future? or...
2. move the `workingPath` option to the `TestCommand` class so it will be better understood as exclusively applicable to Cucumber testing?

If we do the latter, I will likely change the name of the option to something like `testExecutionContextPath` to match the language we decided to use in the configs.  However, if we keep the former behavior, it's probably better to keep it general.

adamlacoste: Per our conversation during a daily status meeting, I will move the option to `TestCommand` and rename it to `testExecutionContextPath`.

- [x] Move context path option to `TestCommand`

wild-love (approved):

PR #277, opened 2026-03-09
Non-Command Input

wild-love:


features/plugin-cli/steps/shell.feature
```diff
@@ -112,7 +112,28 @@ Feature: Shell Step Definitions
...
+            Given that the "@specify-bdd/specify" NPM package is installed
+            And that the "@specify-bdd/plugin-cli" NPM package is installed
+            And a CLI shell
+
+        Scenario: Press a key to respond to a prompt
+            When the user starts the async command "./test/scripts/sh-prompt-key.sh"
+            And waits for the prompt "Press test key"
+            And presses the "Space" key
+            And waits for terminal output matching "Key pressed"
+            Then the last command's terminal output should match "Key pressed: Space"
```
wild-love, line 128: I mentioned this in chat, but I think the redundancy here could be reduced by making the `Then` step async and being an implicit wait.

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
```diff
@@ -139,6 +140,10 @@ defineStep(
     verifyNoMatchingOutput,
 );
 
+defineStep(["When [I press/the user presses] the {string} key"], sendKeyPressToCLI);
+
+defineStep(["When [I input/I enter/the user inputs/the user enters] {string}"], sendLineToCLI);
```
wild-love: Is this the right way to handle multiple versions of "I/the user" phrasing? All of these variations work, but I wasn't sure if we had a convention around it.

features/plugin-cli/steps/shell.feature
```diff
@@ -112,7 +112,28 @@ Feature: Shell Step Definitions
             When the user starts the async command "echo foo; echo bar >&2; sleep 2; echo baz >&2"
             And waits for terminal output on STDERR matching "baz"
             Then the last command's terminal output should match "^foo\nbar\nbaz$"
+    
+    Rule: I can enter non-command input
 
+        Background:
+            Given that the "@specify-bdd/specify" NPM package is installed
+            And that the "@specify-bdd/plugin-cli" NPM package is installed
```
adamlacoste: This problem is already present in the existing shell.feature file so you're welcome to consider this out of scope for right now, but I think it's silly that we're repeating these background steps for pretty much every rule when they should apply to the full file.  There are no scenarios in this feature which don't depend on both specify and plugin-cli packages being installed, since that's what we're testing.  I would create a feature-level background section with these two steps and remove them from the rules background sections.  Many rules will still need a "Given a CLI shell" background step, which can't be moved up to the top because it's not a given in all cases.

adamlacoste: But as I said, if you feel this change should be out of scope, I can create a cleanup ticket for us to circle back to later.

features/plugin-cli/steps/shell.feature
adamlacoste: I agree that some implicit wait behavior should be implemented, but I am concerned that this one risks non-trivial scope creep because we would presumably want all our other Then steps expecting a certain state to have a little elasticity too.  It would feel strange and arbitrary if only some steps behave that way while others are instantaneous checks.

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
adamlacoste: This will work but I think it's a little confusing to read.  Personally I would make this into two patterns.
* `When [I input/the user inputs] {string}`
* `When [I enter/the user enters] {string}`

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
```diff
@@ -230,6 +235,56 @@ async function killCLIShellBySelector(selector?: number | string): Promise<void>
...
+ *
+ * @param line - the line of input to enter
+ */
+function sendLineToCLI(line: string): void {
+    assert.ok(this.cli.manager, new AssertionError({ "message": "No shell session initialized." }));
+
+    line.split("").forEach((char) => sendKeyPressToCLI.call(this, char));
+
+    sendKeyPressToCLI.call(this, "\n");
+}
```
adamlacoste, line 327: Nicely implemented.

modules/@specify-bdd/plugin-cli/src/lib/ShellSession.ts
```diff
@@ -83,12 +83,13 @@ export class ShellSession implements SystemIOSession {
...
      *
-     * @param command - the command to execute. `\n` is appended to simulate a
-     *                  user pressing the "enter" key on their keyboard
+     * @param input         - the input to write.
+     * @param appendNewline - simulate a user pressing the "enter" key on their keyboard
      */
-    write(command: string): void {
-        this.#childProcess.stdin.write(`${command}\n`);
+    write(input: string, appendNewline: boolean = true): void {
+        this.#childProcess.stdin.write(`${input}${appendNewline ? "\n" : ""}`);
```
adamlacoste: I'd strongly prefer if we keep logic like this outside of interpolated strings.  I think a better pattern would be to predetermine a suffix string.

```
const end: string = appendNewline ? "\n" : "";

this.#childProcess.stdin.write(`${input}${end}`);
```

adamlacoste (changes requested):

features/plugin-cli/steps/shell.feature
wild-love: I agree, and I actually almost made this change before creating this PR 😅 I think it is a bit out of scope, but its not a big deal since its a simple refactor in one file. Ill go ahead and do it as part of this work.

- [x] Improve feature background steps

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
wild-love: - [x] Separate into two step defs

modules/@specify-bdd/plugin-cli/src/lib/ShellSession.ts
wild-love: - [x] Predetermine suffix

modules/@specify-bdd/plugin-cli/src/lib/tests/SessionManager.test.ts
```diff
@@ -512,7 +512,7 @@ describe("SessionManager", () => {
             });
         });
 
-        describe.only("sendInput()", () => {
```
adamlacoste, line 515: Oops!  Good catch, I missed this 😅

adamlacoste (approved):
Looks good!

PR #284, opened 2026-03-17
Elastic `Then`'s

wild-love:


features/plugin-cli/steps/shell.feature
wild-love, line 1: We no longer need these types of checks!

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/file_system.ts
```diff
@@ -83,55 +82,51 @@ async function deleteFile(filePath: string): Promise<void> {
...
  * If the file content doesn't match the pattern.
  */
 async function verifyFileContent(filePath: string, pattern: string): Promise<void> {
     await verifyFilePathExists.call(this, filePath);
 
-    const content = await readFile(filePath, "utf8");
-
-    assert.match(content, RegExp(pattern), "File content does not match the expected pattern.");
+    await this.waitFor(async () => RegExp(pattern).test(await readFile(filePath, "utf8")), {
+        "error": Error("File content does not match the expected pattern."),
```
wild-love, line 95: As part of this change, we unfortunately lose the nice default `AssertionError` formatting, so we have to recreate it ourselves. We could make our own `Error` class and make it easier to have some nice formatting, but I thought I'd get some opinions before I dive into that.

modules/@specify-bdd/specify/scripts/generate-types.ts
wild-love, line 1: The changes here and the ones that follow regarding `/types` were made to deal with a strange TS aliasing issue, where `./types/index.d.ts` was not being included in `/dist`, so files modules importing `~/types` were not able to resolve correctly.

I have no idea how this hasn't already failed before now, but these changes seem to resolve the issue for me.

modules/@specify-bdd/specify/scripts/generate-types.ts
```diff
@@ -12,7 +12,7 @@ import { globbySync    } from "globby";
 import type { InterfaceExportOptions } from "./utils";
 
 const __dirname  = import.meta.dirname;
-const outputFile = path.join(__dirname, "..", "types", "index.d.ts");
+const outputFile = path.join(__dirname, "..", "src", "types", "index.ts");
```
wild-love, line 15: Modified to `.ts` so the build step automatically includes it in `/dist`.

modules/@specify-bdd/specify/src/types/index.ts
wild-love, line 1: `/types` was moved to `/src` to comply with `tsconfig.json`'s `rootDir` setting, which is where the build process expects it to be.

modules/@specify-bdd/specify/src/lib/waitFor.ts
```diff
@@ -0,0 +1,88 @@
...
+
+        await new Promise((resolve) => setTimeout(resolve, interval));
+    }
+}
+
+/**
+ * Options for `waitFor(...)` that control polling and timeout behavior,
+ * as well as the error thrown on timeout.
+ */
+export interface WaitForOptions {
```
wild-love, line 68: This is the first interface that has detailed documentation that Im aware of. We should discuss and settle on a convention for these types of docs.

modules/@specify-bdd/specify/src/lib/waitFor.ts
```diff
@@ -0,0 +1,88 @@
+/**
+ * Waits for a predicate function to return true, retrying at specified intervals
+ * until the timeout is reached.
+ *
+ * @param predicate - the function to poll against
+ * @param options   - {@link WaitForOptions}
```
wild-love, line 6: Went with this option from our discussions. It does DRY up the docs a good amount, but in the VSCode intellisense it doesnt show the associated `WaitForOptions` documentation. Instead, it just shows a link to that directs you to the interface's implementation. Is this what we want?

modules/@specify-bdd/specify/src/lib/SpecifyWorld.ts
```diff
@@ -27,6 +32,17 @@ export class SpecifyWorld extends World {
     get quickRef(): QuickRef {
         return this.#quickRef;
     }
+
+    /**
+     * Wrapper for waitFor() with project-specific defaults.
+     * See {@link waitFor} for details
```
wild-love, line 38: Since its the same function signature, I figured Id just use a link to the base function to keep things DRY.

modules/@specify-bdd/specify/src/lib/tests/waitFor.test.ts
```diff
@@ -0,0 +1,216 @@
...
+
+            return true;
+        });
+
+        const expectation = expectToReject(waitFor(fn, { "timeout": 100 }));
+
+        await vi.advanceTimersByTimeAsync(100);
+
+        // will throw a Vitest timeout error if waitFor doesn't timeout correctly
+        await expectation;
```
wild-love, line 204: This isn't my favorite way to do this, but I suppose it works 🤷

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/file_system.ts
adamlacoste: I think our general error handling ought to be a separate ticket.  I agree it needs work, but it's beyond the scope of this ticket.

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/file_system.ts
```diff
@@ -83,55 +82,51 @@ async function deleteFile(filePath: string): Promise<void> {
...
 
-    const content = await readFile(filePath, "utf8");
-
-    assert.ok(
-        !content,
-        new AssertionError({ "message": "Expected file to be empty but it is not." }),
-    );
+    await this.waitFor(async () => !(await readFile(filePath, "utf8")), {
+        "error": Error("Expected file to be empty but it is not."),
+    });
```
adamlacoste, line 115: I don't necessarily think this is wrong, but there's some behavior implication here that I wouldn't necessarily expect.  Because we call the `verifyFilePathExists` handler first, which could theoretically take N seconds to complete if it doesn't exist right away, then we tack on another `waitFor` to verify the content which might take up to N more seconds, we're effectively doubling the timeout here.

Practically speaking, the only way we would be forced to wait the full 2N seconds would be if the file initially didn't exist, then was created with content right before the first timer expired, allowing the second to begin and time out in full.  So it's unlikely to be a problem.  But if we DID want to enforce a strict no-stacking rule on implicit waits, we could wrap both these promises (file exists, file empty) in a `Promise.all` and await _that_.

I genuinely don't know if any behavior change is warranted here, but I wanted to raise the issue for discussion's sake.

modules/@specify-bdd/specify/src/lib/waitFor.ts
adamlacoste: Probably not, but I don't know of a better way to get the result we want apart from repeating the documentation.

modules/@specify-bdd/specify/src/lib/waitFor.ts
adamlacoste: Hm, I don't think we should be documenting the default values on the interface, since those are set somewhere else.  If that should be documented anywhere (I'm not sure it needs to be), it should be on the function that sets them.  IMO the only things we need for each of these properties is a description of what they represent.

adamlacoste (changes requested):
The only thing I would definitely like to see change is the defaultValue documentation on the interfaces.  Everything else, I'd like to discuss but will be happy to approve as is if that's where we land.

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/file_system.ts
wild-love: Good catch! It slipped by me that one `waitFor` handler was calling another. You're right that the behavior might be unintuitive.

Since these are implicit waits instead of explicit, it seems to me that if there are multiple checks, its fine for each to take up to a second (currently) to complete before moving on.

We could alternatively verify the file existence as part of the predicate instead, so we only wait once. Something like;
`await this.waitFor(async () => existsSync(filePath) && !(await readFile(filePath, "utf8")), {`
Which would mean re-implementing parts of some reused handlers.

Alternately redux: Give all assertion step def handlers a `wait = true` argument, and pass false when we use handlers within handlers:  `await verifyFilePathExists.call(this, filePath, false);`

modules/@specify-bdd/specify/src/lib/waitFor.ts
wild-love: I will say one nice thing about documenting the defaults is that they show up when using the options objects.

Image

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/file_system.ts
adamlacoste: Having thought about this some more, I think that your stated principle of allowing each action we take to have its own implicit wait makes a lot of sense.  The overall step timeout is still 5s no matter what implicit waiting we do, so it's not as though we're at risk of letting this run on and on.  And while some logic change to combine the checks (via Promise.all or modifying the predicate) might make a reasonable amount of sense in this case, as a broader principle I think that requiring step defs with multiple actions/checks to always share one implicit wait timeout would be silly.

So ultimately, I think where I've landed is: I'm fine with this as-is, and I think that your approach of making separate implicit waits for each action should be the default way we build these things.  If you want to make any changes to the logic for this specific step def because the file's existence and content are so closely related, I have no objection, but I don't think it's necessary.

modules/@specify-bdd/specify/src/lib/waitFor.ts
adamlacoste: Would you get a similar tooltip if that default value was documented on the function params instead?  IMO that's where it should go.

Ultimately though, the specific tooltip you get is going to be a VSCode behavior.  I don't think we should build the application to cater to that, but I might be biased here because I don't use VSCode.

wild-love: I went ahead and moved it 🤷

adamlacoste (changes requested):
All my feedback was addressed, but I asked Claude to take a look and it found a couple typos and what looks to me like an honest-to-god bug that completely escaped my notice.

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
```diff
@@ -522,19 +520,18 @@ function verifyNoMatchingOutput(pattern: RegExp | string): void {
...
+async function verifyMaximumElapsedTime(maxTime: number): Promise<void> {
     const elapsed = this.cli.manager.commandElapsedTime / 1000;
 
-    assert.ok(
-        elapsed < maxTime,
-        new AssertionError({
-            "message": `The last command's total execution time ${elapsed}s was more than ${maxTime}s.`,
-        }),
-    );
+    await this.waitFor(() => elapsed < maxTime, {
```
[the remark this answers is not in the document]
wild-love: - [ ] Fix the caching error with `elapsed`

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
```diff
@@ -544,19 +541,18 @@ function verifyMaximumElapsedTime(maxTime: number): void {
...
+async function verifyMinimumElapsedTime(minTime: number): Promise<void> {
     const elapsed = this.cli.manager.commandElapsedTime / 1000;
 
-    assert.ok(
-        elapsed > minTime,
-        new AssertionError({
-            "message": `The last command's total execution time ${elapsed}s was less than ${minTime}s.`,
-        }),
-    );
+    await this.waitFor(() => elapsed > minTime, {
```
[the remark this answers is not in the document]
wild-love: - [ ] Dont cache `elapsed`

modules/@specify-bdd/specify/src/lib/waitFor.ts
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * @throws Error
+ * Any error thrown by the predicate
+ */
+export default async function waitFor(
+    predicate: () => boolean | Promise<boolean>,
+    { error = Error("waitFor timed out"), interval = 25, timeout = 1000 }: WaitForOptions = {},
+): Promise<void> {
+    if (interval <= 0) {
+        throw new Error("Invalid waitFor interval, must be a postitive number.");
```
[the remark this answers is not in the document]
wild-love: - [x] Fix typo

modules/@specify-bdd/specify/src/lib/waitFor.ts
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    if (timeout <= 0) {
+        throw new Error("Invalid waitFor timeout, must be a postitive number.");
```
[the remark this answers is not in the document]
wild-love: - [x] Fix typo

adamlacoste (approved):

PR #304, opened 2026-03-30
Revise browser spec

### LINKED ISSUE #303: Revise browser-mode.feature
The feature file at `features/test/browser-mode.feature` was written early in the planning of the Browser Plugin milestone and may not accurately reflect current intentions for how browser mode selection should work. It should be reviewed and updated before any implementation work begins, to ensure it serves as a reliable specification rather than a source of confusion.

## File to review

`features/test/browser-mode.feature`

## Things to consider during review

- Whether the three-mode model (headless / visual / grid) still reflects the intended design
- Whether the CLI flag approach (`--headless`, `--visual`, `--grid <url>`) is still the right mechanism for mode selection, or whether this should be driven by configuration or step definitions instead
- The scenario marked `@todo` ("Grid option defaults to ???") which was left unresolved
- Whether any scenarios should be removed, rewritten, or promoted from `@skip @manual` to automatable

## Why now

This ticket should be resolved before any Browser Plugin step definition or infrastructure tickets are implemented, since those tickets may reference this file as a behavioral specification.

adamlacoste:
Resolves #303.


features/test/browser-mode.feature
adamlacoste, line 1: LMK if anyone objects to this name change.

features/test/browser.feature
```diff
@@ -0,0 +1,98 @@
...
+        And that the "@specify-bdd/plugin-cli" NPM package is installed
+        And a CLI shell
+
+    Rule: Headless mode is the default browser testing mode
+
+        @skip @manual @todo
+        Scenario: Headless testing is the default
+            When the user runs the command "npx specify test ./gherkin/browser/headless.feature"
+            # Then a browser window should not open # redundant with headless testing rule below
+            Then the last command's exit code should be a ${ref.exitCode.success}
```
adamlacoste, line 18: This scenario and the one that follows are marked `@todo` because IMO they are the only two we should be aiming to implement with our Browser plugin MVP.  If our scoping talks revise that, I'll update this.

features/test/browser.feature
```diff
@@ -0,0 +1,98 @@
...
+        And that the "@specify-bdd/plugin-browser" NPM package is installed
+        And that the "@specify-bdd/plugin-cli" NPM package is installed
+        And a CLI shell
+
+    Rule: Headless mode is the default browser testing mode
+
+        @skip @manual @todo
+        Scenario: Headless testing is the default
+            When the user runs the command "npx specify test ./gherkin/browser/headless.feature"
+            # Then a browser window should not open # redundant with headless testing rule below
```
adamlacoste, line 17: I can delete this line if preferred, but since our browser feature is likely to be a WIP for a little while as we build the browser plugin MVP and then iterate on it, I thought it best not to be too hasty, as the context might be helpful.  Eventually though, I think this line should go.

features/test/browser.feature
```diff
@@ -0,0 +1,98 @@
...
+            Then the last command's exit code should be a ${ref.exitCode.success}
+
+    Rule: Headless mode testing is local and invisible
+
+        @skip @manual @todo
+        Scenario: Headless testing is both local and invisible
+            When the user runs the command "npx specify test --headless ./gherkin/browser/headless.feature" # feature stipulates local, non-visible
```
adamlacoste, line 24: This comment is somewhat aspirational, as is the language of the actual headless.feature (below).  I don't yet know how we'll test these things reliably, but that's a later problem.  For now, I just wanted to make sure the intention was stated clearly.  We can remove these comments once we settle on an implementation strategy and no longer need them to explain the intent of what are, for now, placeholders.

features/test/browser.feature
```diff
@@ -0,0 +1,98 @@
...
+
+        @skip @manual
+        Scenario: Headless testing fails if tests must run remotely
+            When the user runs the command "npx specify test --headless ./gherkin/browser/grid.feature" # feature stipulates remote, non-visible
+            Then the last command's exit code should be a ${ref.exitCode.failure}
+
+        @skip @manual
+        Scenario: Headless testing fails if tests must run visibly
+            When the user runs the command "npx specify test --headless ./gherkin/browser/visual.feature" # feature stipulates local, visible
+            Then the last command's exit code should be a ${ref.exitCode.failure}
```
adamlacoste, line 35: My approach here was to test each of the three browser modes (grid, headless, and visual) positively against a feature designed to exercise that mode's distinctions, then negatively against the two other modes' corresponding features, since each would have a requirement the mode being tested could be expected to fail.  While this is a fairly thorough approach, we might find it to be overkill later if these browser tests noticeably extend the total execution time.  If that happens, we could probably live with just the three positive tests of each mode against its corresponding feature file, and axe the negative testing.  For now I'd rather test too much than not enough.

test/gherkin/browser/grid.feature
```diff
@@ -0,0 +1,11 @@
+@browser
+Feature: Grid Browser Testing
+
+    Background:
+        # Given that a Selenium Grid is available at "http://localhost:4444" # this should be handled by the outer spec
```
adamlacoste, line 5: Originally I wrote this feature with a background step ensuring the grid was available, before remembering that I had already done this in the outer browser.feature file (see line 57).  We'll eventually want to delete this line, but I figured I might not be the only person to (at least momentarily) think it was needed, so I left it in for now with a comment.

features/test/browser.feature
```diff
@@ -0,0 +1,98 @@
...
+
+        @skip @manual @todo
+        Scenario: Headless testing is both local and invisible
+            When the user runs the command "npx specify test --headless ./gherkin/browser/headless.feature" # feature stipulates local, non-visible
+            Then the last command's exit code should be a ${ref.exitCode.success}
+
+        @skip @manual
+        Scenario: Headless testing fails if tests must run remotely
+            When the user runs the command "npx specify test --headless ./gherkin/browser/grid.feature" # feature stipulates remote, non-visible
+            Then the last command's exit code should be a ${ref.exitCode.failure}
```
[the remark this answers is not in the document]
adamlacoste: It's intentional.  Specify distinguishes between failing test states and the inability to test, which we feel is important.  I have verified that `exitCode.failure` and `exitCode.error` are both defined in the project refs file.

features/test/browser.feature
```diff
@@ -0,0 +1,98 @@
...
+
+        @skip @manual
+        Scenario: Grid testing fails if tests must run visibly
+            When the user runs the command "npx specify test --grid http://localhost:4444 ./gherkin/browser/visual.feature"
+            Then the last command's exit code should be a ${ref.exitCode.failure}
+
+        @skip @manual
+        Scenario: Grid option defaults to localhost:4444
+            When the user runs the command "npx specify test --grid -- ./gherkin/browser/grid.feature"
```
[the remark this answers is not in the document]
adamlacoste: This usage is how we want the functionality to work; we'll have to ensure during implementation that it does what we intend for it to do with these options.

This does raise a good point though, indirectly: we don't have any scenarios explicitly testing this `--` end-of-options separator functionality.  Maybe we should?

test/gherkin/browser/headless.feature
```diff
@@ -0,0 +1,10 @@
+@browser
+Feature: Headless Browser Testing
+
+    Background:
+        Given that I have a "chrome" browser open
+
+    @todo
+    Scenario: Browser is headless
+        Then the browser window should not be visible # but how can we know?
```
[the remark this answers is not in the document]
adamlacoste: I put the `@todo` tag on scenarios that needed explicit attention when implementing the basic browser testing functionality.  It's OK if they are skipped for now, that is working as intended; because no browser testing functionality exists yet, these tests would fail if they ran and we would be unable to get past our automated PR checks workflow.  As part of the feature that implements the basic functionality, these `@todo` tags should be removed.  I'll make a note of that when I create the ticket for it.

features/test/browser.feature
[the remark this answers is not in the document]
adamlacoste: Claude and I created #305 to verify the behavior of the double dash separator, but it's going to be a challenge to test because we don't currently support any options with optional values.  (For example, you can use `--tags "value"` but not just `--tags`.)  This is intentional, because optional values create parsing ambiguity that can be difficult to resolve.  We wanted all of our options to be clear-cut.

Commander does support `--` end-of-options parsing, so a test verifying this would probably pass if we had an optional-value option to test it with; we just don't right now.  We will if we implement `--grid` as specced here, but it's worth asking if we should, since this would be our first option of that kind, a thing we deliberately avoided before.

I think `--grid` has the best argument yet for supporting that flexibility; you might want grid mode but to use the default (or configured) grid setting, and in that case being required to type out a URL as part of your test command is a drag.  But it would be the black sheep of the family, so we should think about it carefully.

PR #313, opened 2026-04-07
intstr supremacy

### LINKED ISSUE #311: String alternatives for int params
Step definition patterns that use the `{int}` Cucumber parameter type match unquoted numerals (e.g. `exit code should be 0`) but not quoted ones (e.g. `exit code should be "0"`). Since most other step definitions use `{string}` and require quotation marks, the inconsistency can be confusing for test authors who habitually quote their values.

## Work to be done

For each step definition in `shell.ts` that currently accepts an `{int}` parameter, add a corresponding `{string}` pattern alongside the existing one. Update the handler function to accept either a `number` or a `string`, and when the value is a string, convert it with `parseInt`. If `parseInt` returns `NaN`, throw an error with a clear message indicating that the value is not a valid integer.

The affected step definitions are all in `modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts`. The implementer should audit that file for all `{int}` usages and apply this change consistently, unless there is a specific reason a particular step def should not accept a quoted value.

## Example

The exit code step currently registers four patterns using `{int}` and `{ref}`. After this change it should also register the equivalent `{string}` patterns, and `verifyExitCode` should handle both `number` and `string` inputs.

wild-love:
These definitions use the same handler;
```
Then the last command's exit code should be a 1
Then the last command's exit code should be a "1"
```
Accomplished via our brand new param type, ~~`flexInt`~~ `intstr`
```
"Then the last command's exit code/status should be {int}"
"Then the last command's exit code/status should be {intstr}"
```

### DIFF
--- a/features/plugin-cli/steps/shell.feature
+++ b/features/plugin-cli/steps/shell.feature
@@ -16,7 +16,7 @@ Feature: Shell Step Definitions
         Scenario: Create multiple shells
             Given a CLI shell
             When the user starts another CLI shell
-            Then there should be "2" active CLI shells
+            Then there should be 2 active CLI shells
 
         Scenario: Destroy the active shell
             Given a CLI shell
--- a/features/plugin-cli/steps/shell.feature
+++ b/features/plugin-cli/steps/shell.feature
@@ -16,7 +16,7 @@ Feature: Shell Step Definitions
         Scenario: Create multiple shells
             Given a CLI shell
             When the user starts another CLI shell
-            Then there should be 2 active CLI shells
+            Then there should be "2" active CLI shells
 
         Scenario: Destroy the active shell
             Given a CLI shell
--- a/modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
+++ b/modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
@@ -13,18 +13,18 @@ import { ShellSession             } from "@/lib/ShellSession";
 
 import type { SpawnOptions } from "node:child_process";
 
-defineStep("When [I kill/the user kills] CLI shell {int}", killCLIShellByIndex);
+defineStep("When [I kill/the user kills] CLI shell {flexInt}", killCLIShellByIndex);
 defineStep("When [I kill/the user kills] the CLI shell", killCLIShell);
 defineStep("When [I kill/the user kills] the CLI shell named {string}", killCLIShellByName);
 defineStep("When [I send/the user sends] a {cliSignal} signal to the last command", sendKillSignal);
 defineStep("When [I start/the user starts] a/an/the (async )command/process {refstr}", execCommand);
 defineStep("When [I switch/the user switches] CLI shells", switchToNextShell);
-defineStep("When [I switch/the user switches] to CLI shell {int}", switchShellByIndex);
+defineStep("When [I switch/the user switches] to CLI shell {flexInt}", switchShellByIndex);
 defineStep("When [I switch/the user switches] to the CLI shell named {string}", switchShellByName);
 defineStep("When [I wait/the user waits] for terminal output", waitForAnyOutput, {
     "timeout": 60000,
 });
-defineStep("Then there should be {int} active CLI shell(s)", verifyShellCount);
+defineStep("Then there should be {flexInt} active CLI shell(s)", verifyShellCount);
 
 defineStep(
     ["Given a/another CLI shell", "When [I start/the user starts] a/another CLI shell"],
@@ -78,9 +78,9 @@ defineStep(
 
 defineStep(
     [
-        "Then the last command's exit code/status should be {int}",
+        "Then the last command's exit code/status should be {flexInt}",
         "Then the last command's exit code/status should be {ref}",
-        "Then the last command's exit code/status should be a/an {int}",
+        "Then the last command's exit code/status should be a/an {flexInt}",
         "Then the last command's exit code/status should be a/an {ref}",
     ],
     verifyExitCode,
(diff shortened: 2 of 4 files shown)


features/plugin-cli/steps/shell.feature
```diff
@@ -16,7 +16,7 @@ Feature: Shell Step Definitions
         Scenario: Create multiple shells
             Given a CLI shell
             When the user starts another CLI shell
-            Then there should be 2 active CLI shells
+            Then there should be "2" active CLI shells
```
adamlacoste: I don't think we should be quoting our `int` params unless we are explicitly testing the custom param type parsing, which should have its own scenarios built for that purpose.

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
```diff
@@ -13,18 +13,18 @@ import { ShellSession             } from "@/lib/ShellSession";
 
 import type { SpawnOptions } from "node:child_process";
 
-defineStep("When [I kill/the user kills] CLI shell {int}", killCLIShellByIndex);
+defineStep("When [I kill/the user kills] CLI shell {flexInt}", killCLIShellByIndex);
```
adamlacoste: While this approach has the advantage of brevity, I have real concerns about clarity.  Without looking up the definition of `flexInt` (which I don't think we're documenting anywhere currently, nor are we doing that for any of our other custom param types), it's not obvious to me that a typical user would understand what this pattern can match.  The name itself seems deliberately ambiguous, so it's hard to guess how much "flexibility" it has.

My preferred solution would be to use more patterns with narrower types:
```
defineStep(
    [
        "When [I kill/the user kills] CLI shell {int}", // just accepts unquoted integers
        "When [I kill/the user kills] CLI shell {intstr}", // just accepts quoted integers, nothing else
        "When [I kill/the user kills] CLI shell {ref}", // just looks up a ref by name and plugs in the value
    ],
    killCLIShellByIndex,
);
```

That's much longer, I acknowledge, but IMO so much less ambiguous.  The advantage of this approach is that instead of accepting basically any quoted input and determining whether it's a valid integer at the transform stage, we only match against very specific patterns, which leaves us more leeway to define additional patterns which are distinguished by their param formats.

(FWIW, I didn't capitalize the "S" in `intstr` because we have a currently supported param type called `refstr` which you can see used 4 lines below this.  If the capitalization was the only point of contention here, I would be happy to revise `refstr` to use camel case consistently.

adamlacoste (commented):
I have a different approach I want to propose, but if I don't persuade anyone, there might not be any changes needed.

features/plugin-cli/steps/shell.feature
wild-love: - [x] Revert this ~~and add a purpose-build scenario~~

modules/@specify-bdd/plugin-cli/src/cucumber/step_definitions/shell.ts
wild-love: My solution feels cooler, but I can understand your point of view. And on the plus side, each sibling step def string makes it clear we're accepting multiple types of input. 

I'll implement your solution, and I think we should consider adding additional parsing to `defineStep` so we can do the objectively correct solution: `"When [I kill/the user kills] CLI shell {int|intstr|ref}"`

wild-love: - [x] Refactor `flexInt` as `intstr`

features/plugin-cli/steps/shell.feature
wild-love: I decided to just revert the change. I manually tested the param type, and since we don't test other param types via scenarios, I figured I'd wait to do so until we make a decision on that about them all.

modules/@specify-bdd/plugin-cli/src/cucumber/support/param_types.ts
```diff
@@ -9,6 +9,9 @@ import { defineParamType        } from "@specify-bdd/specify";
 import assert, { AssertionError } from "node:assert/strict";
 import { constants              } from "node:os";
 
+// matches "1", "-2", '3', but not '4". Auto-strips quotes
+const quotedInt = /'(-?\d+)'|"(-?\d+)"/;
```
wild-love: This pattern causes cucumber to treat each side of the matcher individually, and it returns a `string[]`. Not my favorite way to do it, but it seemed like the best of the bad options 🤷 This necessitated a change in `CucumberManager` to conform to the base Cucumber implementation, seen below.

modules/@specify-bdd/specify/src/lib/CucumberManager.ts
```diff
@@ -34,7 +34,7 @@ export interface ParamTypeOptions {
     name: string;
     preferForRegexpMatch?: boolean;
     regexp: RegExp;
-    transformer?: (arg: string) => any; // eslint-disable-line @typescript-eslint/no-explicit-any
+    transformer?: (...args: string[]) => any; // eslint-disable-line @typescript-eslint/no-explicit-any
```
wild-love, line 37: I had to make this a variadric signature, which matches how cucumber does it. The code would actually work without this change, but TS complains about it in the editor.

modules/@specify-bdd/plugin-cli/src/cucumber/support/param_types.ts
adamlacoste: Couldn't we achieve the same thing with a pattern like `/'-?\d+'|"-?\d+"/` and then just strip the quotes off in the transform function?  It's not that I don't think your solution would work -- it's pretty clever, in fact -- but it seems more complicated than it needs to be, and is inconsistent with how we handle our normal `quotedString` patterns where we use `input.slice(1, -1)` like simple cavemen.  My suspicion here is that simpler might be better.

adamlacoste (changes requested):

modules/@specify-bdd/plugin-cli/src/cucumber/support/param_types.ts
wild-love: - [ ] Use the existing `.slice()` pattern

adamlacoste (approved):

